PR #93, opened 2024-08-13
Simplify resource configuration

kels271828:
- To test the pipeline, set `cluster_name='dummy'`
- To run the pipeline, set `cluster_name='slurm` (any valid name other than 'dummy')
  - Resources file at /directory/config/resources.yml required
- Removed `configure_resources` argument
- Renamed `resource_path` to `resource_yaml`

### DIFF
--- a/src/onemod/main.py
+++ b/src/onemod/main.py
@@ -12,17 +12,17 @@ except ImportError:
     pass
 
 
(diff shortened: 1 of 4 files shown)


geoffreyphipps-uw (approved):
looks good

PR #100, opened 2024-09-25
Updated package design

kels271828:
The main thing to review in this PR is the new file structure.


src/onemod/stage/model_stage/kreg_stage.py
```diff
@@ -0,0 +1,46 @@
+"""Kreg stage."""
+
+import fire
+
+from onemod.config.model_config import KregConfig
+from onemod.stage import ModelStage
+
+
+class KregStage(ModelStage):
```
blsmxiu47, line 9: The way these stage classes are written means we can immediately write simple tests for the construction and running of a dummy Pipeline 👍 which is not a bad place to get things started I think

src/onemod/pipeline.py
```diff
@@ -0,0 +1,144 @@
...
+                stage.create_stage_params()
+        self._stages[stage.name] = stage
+
+    def _stage_from_json(self, stage: str, filepath: Path | str) -> Stage:
+        """Load stage object from JSON."""
+        if stage in self.stages:
+            raise ValueError(f"stage '{stage}' already exists")
+        with open(filepath, "r") as f:
+            stage_json = json.load(f)
+        if stage_type := stage_json["type"] in STAGE_DICT:
```
blsmxiu47, line 99: I might need some convincing on the use of `STAGE_DICT` here. I'm not exactly sure what a best alternative looks like at this moment, but for some reason I feel like within the core Pipeline component there might be a nicer alternative to relying on a static dictionary of stage types. Might be missing something though

blsmxiu47:
Thought on structure/naming - I wonder if we could make the names of directories which hold classes for implementations on top of the base `Stage` classes plural (i.e. `model_stage` -> `model_stages`, etc.). Tiny thing but to me that makes it a bit clearer that these are instances of Stages built on top of `ModelStage` (etc) rather than components required in the core implementation of `ModelStage` (etc)

src/onemod/stage/model_stage/kreg_stage.py
geoffreyphipps-uw: I think it is a good place to start :-)

blsmxiu47 (approved):
This will serve as a great base for us to work off of. Thanks for your hard work!

src/onemod/pipeline.py
kels271828: Agreed, removed for now; we can figure out an easier way to load built-in stages later

PR #101, opened 2024-09-30
bugfix/json_funcs

kels271828:
Fixed some minor issues to get code to run, then made changes to the stage `module` property and how pipelines and stages are loaded from JSON. Previously, `module` came from `inspect.getfile(self.__class__)`, but this didn't work if the stage was created by `Pipeline.from_json` (e.g., calling `my_stage.module`, `my_pipeline.to_json()`, or `my_stage.to_json()` gave error messages like "TypeError: <class 'my_stage.PreprocessingStage'> is a built-in class").

**Changes made**

Minor changes
* Added `PipelineConfig`, `PreprocessingConfig`, `KregConfig`, `RoverConfig`, `SpxmodConfig` to config/__init__.py
* Removed "data_config" and "model_config" from import statements in preprocessing_config.py, kreg_config.py, rover_config.py, and spxmod_config.py
* Added `PreprocessingStage`, `KregStage`, `RoverStage`, and `SpxmodStage` to stage.__init__.py
* Removed "redesign" from import statements in data_config.__init__.py and preprocessing_config.py
* Replaced `PreprocessingConfig` import with `PreprocessingStage` in data_config.__init__.py
* Changed return type of `Pipeline.stages` property from `set[str]` to `dict[str, Stage]`
* Add `self.name` to `Stage.run()` methods (this doesn't really matter since the methods will get rewritten when we actually implement the stages)
* Changed `Pipeline.add_stage()` and `Pipeline.add_stages()` to only accept `Stage` objects

`module` property
* OneMod stages: `None`
* Custom stages: `self._module` if created by `Pipeline.from_json()`, `inspect.getfile(self.__class__)` otherwise

`from_json` methods
* Added `from_pipeline` argument to `Stage.from_json()` and `Stage.evaluate()`, so the functions know if they are working with a pipeline or a stage config file
* Synced the default arguments for `name` and `from_pipeline` in `Stage.from_json()` and `Stage.evaluate()` (i.e., `name` is required if `from_pipeline` is True, so set `from_pipeline` default to False)
* Added `stage.module` assignment to `Stage.from_json()`
* Added different way to load OneMod stages in `Pipeline.from_json()` (i.e., get class directly from `onemod.stage` rather than from module path)

**Instruction for reviewers**

Does the handling of the stage `module` property make sense?

### DIFF
--- a/src/onemod/stage/base.py
+++ b/src/onemod/stage/base.py
@@ -53,12 +53,23 @@ class Stage(BaseModel, ABC):
         return self._skip_if
 
     @classmethod
-    def from_json(cls, filepath: Path | str, name: str) -> Stage:
+    def from_json(
+        cls,
+        filepath: Path | str,
+        name: str | None = None,
+        from_pipeline: bool = False,
+    ) -> Stage:
         """Create stage object from JSON file."""
-        with open(filepath, "r") as f:
-            pipeline_json = json.load(f)
-        stage = cls(**pipeline_json["stages"][name])
-        stage.directory = pipeline_json["directory"]
+        if from_pipeline:
+            with open(filepath, "r") as f:
+                pipeline_json = json.load(f)
+            stage = cls(**pipeline_json["stages"][name])
+            stage.directory = pipeline_json["directory"]
+        else:
+            with open(filepath, "r") as f:
+                stage_json = json.load(f)
+            stage = cls(**stage_json)
+            stage.directory = Path(filepath).parent
         return stage
 
     def to_json(self, filepath: Path | str | None = None) -> None:
--- a/src/onemod/stage/base.py
+++ b/src/onemod/stage/base.py
@@ -11,6 +11,7 @@ from typing import Any
 from pandas import DataFrame
 from pydantic import BaseModel, ConfigDict, computed_field
 
+import onemod.stage as onemod_stages
 from onemod.config import StageConfig, GroupedConfig, CrossedConfig, ModelConfig
 from onemod.utils.parameters import create_params, get_params
 from onemod.utils.subsets import create_subsets, get_subset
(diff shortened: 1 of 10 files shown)


src/onemod/stage/base.py
```diff
@@ -53,41 +66,105 @@ def skip_if(self) -> set[str]:
         return self._skip_if
 
     @classmethod
-    def from_json(cls, filepath: Path | str, name: str) -> Stage:
-        """Create stage object from JSON file."""
+    def from_json(
+        cls,
+        filepath: Path | str,
+        name: str | None = None,
+        from_pipeline: bool = False,
```
blsmxiu47, line 73: Just to make sure I am understanding the use cases here. The use case for this `from_pipeline` argument could be if someone wants to use a single json config for their Pipeline (makes sense), but they need to initialize and/or evaluate some portion of their stages separately. Is this correct? Now that I type it out I think it probably is desirable to have the option.

blsmxiu47:
re: does handling of `module` property make sense - Here `module` does not have any particular OneMod-specific meaning right (if so, I think this is good because otherwise we might start conflating terms)? It is just that we need the ability to import Python modules, i.e. custom stages, as valid OneMod Stages, so this property would point to the location of that custom stage, if not a OneMod built-in Stage class, correct?

There may be further discussion warranted around different ways to import or otherwise point to custom Stages, but I would say that for now it makes sense as long as it functions as expected, in context. I.e we may confirm whether it makes sense when we actually try out importing a custom stage in a test pipeline.

kels271828:
> re: does handling of `module` property make sense - 

Yes, `module` is only used for custom stages. Since I'm assuming the user won't modify the OneMod package to include their custom stage, `module` is the path to where their stage class is defined so OneMod can load it (required for the `Pipeline.from_json` method, but now that I'm thinking about it, potentially also required for Jobmon tasks, even for OneMod stages, if we don't use entrypoints like the main branch). Open to other name suggestions, or even requiring the user to save their definition in a specific place with a specific naming convention (but this seems potentially more confusing for the user).

This was just my first approach for handling custom stages, so we can definitely find a better way if this one doesn't work out.

src/onemod/stage/base.py
blsmxiu47: sgtm. I think it makes sense to leave the default as `False`

PR #102, opened 2024-09-30
Initial Pipeline DAG validation and testing

blsmxiu47:
## Description

Quick PR to add some general DAG validation functionality to the Pipeline class and cleanup the testing directory. Some of these changes will likely be soon obsoleted by upcoming tasks, namely how the assignment of `Pipeline` inter-stage dependencies works from a syntax perspective, however this update may still be useful to include for the following reasons:

- Addition of first redesign-related testing examples
- All or parts of `validate_dag()` and `get_execution_order()` may possibly be regardless of how dependencies are first assigned
- Until we finalize stage dependency assignment, the simple argument-based method can be used in the mean time while we test out dag validation and compilation

## Changes Made

- Add `Pipeline.validate_dag()`, `Pipeline.get_execution_order()`
- Allow dependencies list to be passed to `add_stage()` for now
- Add related test cases for the above
- Remove `tests/` sub-directories corresponding to pre-redesign modules

## Additional Notes

- Branch name is a bit dated. MSCA-306 ended up becoming a broader design ticket that got split into several implementation tasks; this PR touches on only the part of it described above.


src/onemod/pipeline.py
```diff
@@ -34,12 +38,18 @@ class Pipeline(BaseModel):
     data: Path | None = None
     groupby: set[str] = set()
     _stages: dict[str, Stage] = {}  # set by Pipeline.add_stage
+    _dependencies: dict[str, list[str]] = {}
```
kels271828, line 41: Could we simplify some of the validation by using a set instead of a list?

src/onemod/pipeline.py
```diff
@@ -175,6 +209,96 @@ def stage_from_json(
...
+                raise ValueError(f"Stage '{stage}' cannot depend on itself.")
+
+            # Check for duplicate dependencies
+            if len(dependencies) != len(set(dependencies)):
+                raise ValueError(
+                    f"Duplicate dependencies found for stage '{stage}'."
+                )
+
+        # Check for isolated nodes
+        # TODO: are there cases where isolated nodes would be valid for OneMod?
```
kels271828, line 238: I can't think of any cases

src/onemod/pipeline.py
```diff
@@ -175,6 +209,96 @@ def stage_from_json(
...
+        all_stages = set(self._stages.keys())
+        dependent_stages = set(
+            stage for deps in self._dependencies.values() for stage in deps
+        )
+        isolated_stages = (
+            all_stages - dependent_stages - set(self._dependencies.keys())
```
kels271828, line 244: I think every stage gets added to both self._stages and self._dependencies (possibly with an empty list) when added to the pipeline, so you might have to filter the last term by stages that actually have dependencies.

kels271828: Assuming this is all_stages - (stages that other stages depend on) - (stages that depend on other stages)

src/onemod/pipeline.py
```diff
@@ -175,6 +209,96 @@ def stage_from_json(
...
+        Return topologically sorted order of stages, ensuring no cycles.
+        Uses of Kahn's algorithm to find the topological order of the stages.
+        """
+        # Track in-degrees of all nodes
+        in_degree = {stage: 0 for stage in self._stages}
+
+        # Populate in-degree based on dependencies
+        for stage, dependencies in self._dependencies.items():
+            for dep in dependencies:
+                in_degree[stage] += 1
```
kels271828, line 262: Should this be `in_degree[dep]`?

kels271828: Oops, no, the in-degree is the number of incoming nodes, i.e., the number of dependencies a node has, right? Could we use `in_degree = {stage: len(dependencies) for stage, dependencies in self._dependencies.items()}` in line 257? I'm assuming all stages are in `self._dependencies`. If not, could we use `in_degree[stage] = len(dependencies)` in line 261?

kels271828 (approved):

src/onemod/pipeline.py
blsmxiu47: Yes. Actually that was my initial thought and implementation, but ultimately thought maybe the target user would be more comfortable with list syntax, at the expense of an extra line of validation. Also was thinking the size of the list of dependencies would not be large enough that we would realize major efficiency improvements.

However, since users won't actually specify dependencies by list going forward, if we do have the concept of stage dependencies under the hood it should probably be set now.

src/onemod/pipeline.py
blsmxiu47: We can. The first suggestion would mean we are relying on `_dependencies` only, rather than initializing the set of stages based on `_stages`, which is probably fine, as that should probably be validated elsewhere (I actually think they may not need to be separate entities ultimately, but this is a larger design question that we are in the process of answering). Either way, will update so the method is less busy.

src/onemod/pipeline.py
```diff
@@ -175,6 +209,96 @@ def stage_from_json(
...
+        rec_stack = set()
+        visited = set()
+
+        while queue:
+            stage = queue.popleft()
+            topological_order.append(stage)
+            visited.add(stage)
+
+            # Reduce the in-degree of dependent stages
+            for dep in self._dependencies.get(stage, []):
```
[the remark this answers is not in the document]
blsmxiu47: Yes, thank you for catching this. Updated to hopefully fix - now constructs a "reverse" graph that includes the downstream dependents for a given stage, and updates in-degree based on these. Resulted in me to re-add the above `for` loop though.

PR #103, opened 2024-10-08
Dependency specification

kels271828:
### Description

Added IO classes and Stage methods to handle stage dependencies.

- `Dummy` class is just a placeholder for new onemod types
- `Input` and `Output` classes keep track of a stage's input and output, specifically filepaths and stage names
- `Input` class provides some validation, including checking for missing input, checking for cycles (can only check if there is a self-dependency), and checking types (we might want to update this part when we rewrite to use new onemod types)
- Instead of creating subclasses for specific stages like we have for the `Config` class, stage classes define their expected input and output via the fields `_required_input`, `_optional_input`, and `_output` and store items in `items` property
- `Input` and `Output` classes implement some dictionary methods to simplify access to items (e.g., `my_stage.output["output_name"]` vs. `my_stage.output.items["output_name"]`)
  - What if we moved these dictionary methods to stage class (e.g., `my_stage["output_name"]` vs. `my_stage.output["output_name"]`?)
- Users can configure stage input either using the `input` keyword in `Stage.__init__` or input item keywords in `Stage.__call__`
- Stage output instances are frozen

### Changes made

* Updated examples to show different ways to set stage dependencies
* Removed input fields from stage config classes (these are now mostly empty, but they will eventually have lots of model settings)
* Added `Data`, `IO`, `Input`, and `Output` classes
* Added `input`, `_required_input`, `_optional_input`, and `_output` fields to `Stage` class
* Added `output` and `dependencies` properties to `Stage` class
* Changed `name` arg to `stage_name` in `Stage` methods because Jobmon doesn't allow `name` args
* Added `__call__` method to `Stage` class
* Added `_required_input`, `_optional_input`, and `_output` fields to data and model stage classes
* Added tests for input, output, and stage input/output


src/onemod/pipeline.py
```diff
@@ -19,26 +19,30 @@
...
+        stage `groupby` attribute. Default is None.
+    groupby : set[str] or None, optional
+        ID names used to create data subsets. Default is None.
 
     """
 
     name: str
     config: PipelineConfig
-    directory: Path
+    directory: Path  # TODO: replace with DataInterface
```
blsmxiu47, line 42: Were you thinking of using `DataInterface` from `pplkit` as an imported module or we should create a module within OneMod that takes inspiration from it?

blsmxiu47 (approved):
Ran unit tests, all passed. ✅ 
I have kinda been following along with these updates already, so no real surprises. Just had one clarifying question.

Otherwise, thanks for the quick PR! ⚡ 
Once this gets merged I'll apply these I/O and Stage updates to my compilation branch. Some things (e.g. test scripts) may get re-organized a bit, and might have more specific questions at that time, but we'll see if we can't get a working (local) e2e prototype here soon! 😄

src/onemod/pipeline.py
kels271828: I was thinking of importing it from pplkit

PR #105, opened 2024-10-11
Stage IO Data, Constraints and Validation

blsmxiu47:
## Description

Request to merge in updates for the `constraints` module and `Data` validation on `Stage` inputs and outputs.

Unfortunately, `Pipeline.build()` and serialization is not completed, but this branch has already taken on far more responsibility than it probably should have and I want to make sure that any updates that we would like to discuss and/or build on from this branch may be integrated so that it's easier for us to work in parallel going forward.

### Examples

- Adds example Pipeline json output. Maybe not final or a full example but this is roughly the goal result for `Pipeline.build()`
- Adds ideas for a common `OneMod` project directory structure. Definitely not final, but something to start from/build upon.

### Constraints

- Adds `constraints` module
- Adds `Constraint` class for defining and registering constraint functions to be applied across Polars Series for data validation.
- Constraints may be passed within `Data` when defining i/o validation
- Adds builtin constraint functions `bounds()`, `is_in()`, `no_inf()`, mostly to serve as examples and for testing purposes

### IO

- Originally updated `Inputs` and `Output` to take `Data` validation directly, but partially updated following https://github.com/ihmeuw-msca/OneMod/pull/103 to reconcile differing design approaches. Reconciliation updates still need to be made here, especially wrt serialization.
- Adds `validate()` for inputs and outputs using `validation_schemas`. May be better ways to handle this.

### Serializers

- Adds `serializers` module
- Adds initial serialization and deserialization functions. These could be greatly expanded if useful across OneMod. Serialization needs to be standardized further throughout if possible

### Stage

- Adds `input_types`, `output_types` to allow passing I/O validation specs
- Adds methods for `validate_inputs()`, `validate_outputs()`

### Types

- Adds `Data` type for validation of data i/o
- Adds `ColumnSpec` type for `Data` column-level validation
- Adds `FilePath` type based on Pydantic's `FilePath`. Validates extension. Functionality could be extended though ultimately may not be too useful in practice.

### Utils

- Adds `DataIOHandler` for reading and writing `Data`. Will need to be greatly expanded, functionality may belong somewhere elee ultimately

### Validation

- Adds `validation` module
- Adds `ValidationErrorCollector`, `ValidationErrorReport` to allow for accumulation of validation errors for batch reporing when running locally with stage i/o `Data` validation.

### Tests

- Reorganization of tests into `unit`, `integration`, `e2e`, though the organization could be cleaned up for sure based on preferences
- Added tests for new functionality, and some tests for desired but not yet complete functionality pending the below items ⬇️ 

## Partially completed items / Open questions

- Reconciling the serialization of IO along with validation has been trickier than expected. It's possible (or maybe probable) there is a very obvious solution and I have just been staring at the same code for too long, so definitely part of the reason for making this PR to get feedback, but either way we still need to standardize serialization methods, which will also unlock `Pipeline.build()` and end-to-end testing.
- `_check_cycles()` for `Input` class: assumes the `stage` key of `Data` type refers to the origin stage,  which may be desirable, but this would need to be adjusted for `Data`, as `Data.stage` still refers to the stage to which the i/o belongs, rather than origin.


examples/project_directory_structure_ideas.md
```diff
@@ -0,0 +1,25 @@
+# OneMod Project Directory Structure
+
+Ideas for how a OneMod project directory structure could be organized. Would be initialized with a command like `onemod init` or `onemod create` or `onemod new` or `onemod startproject` or similar.
```
kels271828, line 3: I vote for `onemod init`

examples/project_directory_structure_ideas.md
```diff
@@ -0,0 +1,25 @@
...
+    config/
+        config.yaml
+    experiments/
+        config/
+            stage1/
+                config.yaml
+            pipeline_config.yaml
+            run_config.yaml
+    scripts/
+        run.py
```
kels271828, line 21: Does this script create the config files?

examples/example_onemod_pipeline_representation.json
```diff
@@ -0,0 +1,153 @@
...
+        ]
+    },
+    "observations_column": "obs",
+    "predictions_column": "pred",
+    "holdout_columns": [
+        "holdout1",
+        "holdout2"
+    ],
+    "test_column": "test",
+    "mtype": "binomial",
```
kels271828, line 37: We could change this to `model_type` to be more descriptive like the other items

src/onemod/pipeline.py
```diff
@@ -262,42 +240,145 @@ def get_execution_order(self) -> list[str]:
...
+        
         Notes
         -----
-        * These functions will handle workflow creation
-        * Maybe allow subsets of the DAG to be run (e.g., predict for a
-          single location)
-        * TODO: run for selected stages
-        # TODO: run for selected subsets or params
-
+        * TODO: Include (?) OneMod, project version info
```
kels271828, line 296: This would be great, we would love suggestions for how we can track versions!

src/onemod/utils/data_io_handler.py
```diff
@@ -0,0 +1,35 @@
+from pathlib import Path
+
+from polars import DataFrame, read_parquet
+
+
+class DataIOHandler:
```
kels271828, line 6: What are your thoughts about writing our own handler vs. using `pplkit.DataInterface`? Besides read/write, `DataInterface` also keeps track of directory structure, but we could create a OneMod version of that too.

src/onemod/types/filepath.py
```diff
@@ -0,0 +1,14 @@
+from pydantic import BaseModel, FilePath as PydanticFilePath, model_validator
+
+
+class FilePath(BaseModel):
+    path: PydanticFilePath
+    extension: str | None = None
+    
+    @model_validator(mode="after")
+    def check_extension(self):
```
kels271828, line 9: `@model_validator` vs. `model_post_init`: Is the advantage that the function name makes it clear what it does? I guess when I've used `model_post_inti`, I was often doing things other than validation.

src/onemod/stage/model_stages/rover_stage.py
```diff
@@ -31,6 +31,12 @@ def fit(
             f"fitting {self.name} submodel: subset {subset_id}, param set {param_id}"
         )
 
+    def predict(
```
kels271828, line 34: Should users write something similar when their stages skip methods?

src/onemod/types/__init__.py
```diff
@@ -0,0 +1,9 @@
+from onemod.types.column_spec import ColumnSpec
+from onemod.types.data import Data
+from onemod.types.filepath import FilePath
```
kels271828, line 3: Where will `FilePath` be used?

src/onemod/serializers/functions.py
```diff
@@ -0,0 +1,44 @@
+import json
+from pathlib import Path
+from typing import Union
```
kels271828, line 3: Not sure if this is just a style choice or there is a true better/worse way of doing things. Opinion on using `Union[str, Path]` vs. `str | Path`, etc.? Or if it's important that the package uses consistent style? Awhile ago Peng removed `Union` in favor of `|`. Similar with `Optional[str]` and `str | None` cases.

src/onemod/serializers/functions.py
```diff
@@ -0,0 +1,44 @@
...
+    
+    with open(file_path, 'w') as file:
+        if file_format == 'json':
+            if isinstance(model, BaseModel):
+                file.write(model.model_dump_json(indent=4, exclude_none=True, serialize_as_any=True))
+            else:
+                json.dump(model, file, indent=4)
+        elif file_format == 'yaml':
+            if isinstance(model, BaseModel):
+                yaml.safe_dump(model.model_dump(), file, default_flow_style=False)
```
kels271828, line 42: We might need `serialize_as_any=True` in `model_dump()` as well. When serializing a pipeline instance including stages, not all items in the stage configs were getting saved. There may be other ways to handle this: https://docs.pydantic.dev/latest/concepts/serialization/#subclass-instances-for-fields-of-basemodel-dataclasses-typeddict

src/onemod/constraints/functions.py
```diff
@@ -0,0 +1,81 @@
...
+        callable: Function that validates the column values
+    """
+    
+    def validate(column: Series) -> None:
+        if not column.is_in(other).all():
+            raise ValueError(f"All values must be in {other}.")
+    
+    return validate
+
+def no_inf() -> callable:
```
kels271828, line 62: I think it could be useful to add a similar `no_nan` validator

src/onemod/constraints/base.py
```diff
@@ -0,0 +1,78 @@
+from typing import Any, Callable, Dict
+from polars import Series
+
+from pydantic import BaseModel, field_validator
+
+from .functions import bounds, is_in
```
kels271828, line 6: What's the difference between importing with `.functions` and `onemod.constraints.functions`?

src/onemod/constraints/base.py
```diff
@@ -0,0 +1,78 @@
...
+        The validation function to apply.
+        
+    Examples
+    --------
+    >>> def custom_constraint_example(limit: int) -> Callable:
+    ...     def validate(column: Series) -> None:
+    ...         if not column.lt(limit).all():
+    ...             raise ValueError(f"Values must be less than {limit}.")
+    ...     return validate
+    >>> register_constraint("custom_constraint", custom_constraint_example)
```
kels271828, line 67: When/where would a user register a new constraint?

src/onemod/stage/base.py
```diff
@@ -127,13 +136,43 @@ def from_json(
             config = config["stages"][name]
         else:
             directory = Path(filepath).parent
-        stage = cls(**config)
+        stage = cls(
```
kels271828, line 139: What's the difference between `**config` and explicitly passing arguments?

src/onemod/stage/base.py
```diff
@@ -149,52 +188,88 @@ def to_json(self, filepath: Path | str | None = None) -> None:
         filepath = filepath or self.directory / (self.name + ".json")
         with open(filepath, "w") as f:
             f.write(self.model_dump_json(indent=4, exclude_none=True))
+            
+    def to_dict(self) -> dict:
```
kels271828, line 192: Is this different from calling `stage.model_dump()`?

src/onemod/stage/base.py
```diff
@@ -149,52 +188,88 @@ def to_json(self, filepath: Path | str | None = None) -> None:
...
-        stage = cls.from_json(filepath, stage_name, from_pipeline)
-        if method in stage.skip_if:
-            raise AttributeError(f"{stage.name} skips the '{method}' method")
+        if method in self.skip_if:
+            raise AttributeError(f"{self.name} skips the '{method}' method")
+        if backend == "jobmon":
+            raise NotImplementedError("Jobmon backend not implemented")
         try:
-            stage.__getattribute__(method)(*args, **kwargs)
+            self.__getattribute__(method)()
```
kels271828, line 235: I think we still need to pass arguments (could be subset_id, param_id), I accidentally remove them but added them back after trying to run things

src/onemod/stage/base.py
```diff
@@ -303,23 +363,7 @@ def set_params(self, param_id: int) -> Any:
...
-        **kwargs,
-    ) -> None:
-        """Evaluate stage method."""
-        stage = cls.from_json(filepath, stage_name)
-        if method in stage.skip_if:
-            raise ValueError(f"invalid method: {method}")
-        stage.__getattribute__(method)(param_id, *args, **kwargs)
-
-    def run(self, param_id: int) -> None:
+    def run(self, subset_id: int | None = None, param_id: int | None = None) -> None:
```
kels271828, line 366: `GroupedStage` doesn't use `param_id` and `CrossedStage` doesn't use `subset_id`, but since we are merging them into `ModelStage` this won't matter

src/onemod/io/base.py
```diff
@@ -72,7 +117,20 @@ def dependencies(self) -> set[str]:
         return set(
             item.stage for item in self.items.values() if isinstance(item, Data)
         )
-
+        
+    @classmethod
+    def from_dict(cls, data_dict: dict) -> 'Input':
```
kels271828, line 122: The way I was thinking of things, `required` and `optional` didn't contain actual data, they just contained the names and extensions of data expected for the stage. They come from the stage definition (i.e., the same across all stage instances). The actual input data for a stage instance would be in the `items` attribute.

src/onemod/io/base.py
```diff
@@ -182,6 +242,32 @@ class Output(IO):
     """Stage output class."""
 
     items: dict[str, Data] = {}  # defined by stage class
+    _expected_names: set[str]
```
kels271828, line 245: For the output class, it would be created by the stage definition and be the same across all instances, so I was thinking that doing checks and serialization seemed unnecessary. But we can talk about it if you've changed how the `Input`/`Output` classes work!

examples/example_onemod_pipeline_representation.json
blsmxiu47: Sounds good

src/onemod/types/__init__.py
blsmxiu47: good point :)

src/onemod/serializers/functions.py
blsmxiu47: No opinion either way, but agree that we should be consistent

src/onemod/constraints/functions.py
```diff
@@ -0,0 +1,81 @@
...
+
+from polars import Series
+
+def bounds(ge: float = None, le: float = None) -> Callable[[Series], None]:
+    """
+    Returns a function that checks if all values in the column are within the specified bounds.
+    
+    Parameters
+    ----------
+    ge (float): Minimum allowed value (inclusive)
```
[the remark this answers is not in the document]
blsmxiu47: These are just a couple examples. Let's definitely add more as useful

src/onemod/constraints/base.py
blsmxiu47: no difference, stylistic thing

kels271828 (approved):

PR #107, opened 2024-10-16
Removed bad quote in pyproject.toml

geoffreyphipps-uw:
Could not install with pip, this fixes it.


kels271828:
Tried to test these changes with Jobmon this morning, but looks like there might be a maintenance event in progress (got DistributorNotAlive exception). Will try again later.

There are various import statements that will also need to be changed from `dtypes` to `types`.

kels271828:
I'm still getting the same error when I run with Jobmon, but running locally is fine. We can pull in the fix for the quote, but I think we still need to figure out the types vs. dtypes issue.

kels271828:
The bad quote was fixed elsewhere, so closing this PR.

PR #108, opened 2024-10-18
More MVP Testing

blsmxiu47:
## Description

Updates/adds to integration tests involving `Pipeline.build()`, adds an e2e test for `Pipeline.evaluate()`, and organizes the testing directory a bit further. Also a few minor fixes to `pyproject.toml`.

### Pipeline

- Added an optional path argument to `Pipeline.build()`. Still defaults to saving config to `self.directory / f"{self.name}.json"`.

### Tests

- Integration tests for `Pipeline.build()`, namely asserting whether a `Pipeline` reconstructed from json (after being saved out by `build()`) matches the original.
- Adds `tests/helpers` directory, with utility functions and a set of "dummy" Stage classes so that we still can have "no-logic" stages for pipeline build/execution tests, even after the actual OneMod Stages are implemented.
- Adds e2e test for sequential (local) build and execution of a dummy pipeline. This passing means the next step is to try a similar exercise with "real" Stages once they are implemented.

### pyproject.toml

- Was testing out local environment setup and noticed there are still some dependencies listed that I believe we don't actually use in the redesign (`loguru`, `matplotlib`, `scipy`, `seaborn`). Please review and correct me if I'm wrong and they still are used somewhere. We can of course add them back in later as needed, but prefer to keep the `pyproject.toml` specific to what is actively used so far.
- Tried using Python 3.13 with OneMod but at least one of our internal dependencies (think it might've been `regmod` that I noticed) doesn't support so updated the `requires-python` versions for now until we can investigate.
- `python-dotenv` is only needed in dev dependencies atm so removed from core dependencies list

## Additional notes

- Have been thinking about the split of testing, i.e. what can be automated with Actions vs. tests that MSCA will need to run manually/tests that may be specific to internal environments. One of my next tasks is related to this and I believe the "heavier" e2e test using realistic data assets, Jobmon backend will come either following or as a part of that task, or else following the implementation of our "real" OneMod stages, whichever comes last
- There are further updates to `pyproject.toml` that I want to make, but this PR is not the place. Just included these ones because it helped me get my environment set up to work on this branch


pyproject.toml
```diff
@@ -21,21 +21,16 @@ classifiers = [
 ]
 dependencies = [
     "fire",
-    "loguru",
```
kels271828, line 24: I've used logging in the rover and spxmod stages I'm working on, but I can add it back when I make my PR. We can add the others back as needed!

kels271828 (approved):

PR #109, opened 2024-10-22
Initial pre-commit

blsmxiu47:
## Description

Adds pre-commit configuration and copies over existing `ruff` configuration into `pyproject.toml`, just to organize everything in one place. If 

## Changes made

- Add `.pre-commit-config.yaml` with ruff, mypy (mirroring msca's `pypkg` repo example)
- Remove `.flake8`
- Move `ruff.toml` configuration into `pyproject.toml`
- Add isort (ruff select += ["I"]) to ruff linting

## Additional notes

Small PR, just let me know if there are any preferences for linting, etc. that we could easily add. Or if we prefer to have the `ruff.toml` be separate from `pyproject.toml` for some reason.


blsmxiu47:
@kels271828 I wanted to check a few things before merging this PR, as it's mostly about preferences:
- Would it be okay to migrate `mypy.ini` configuration into `pyproject.toml`, and/or do we need to keep the dependency-specific `ignore_missing_imports` in [`mypy.ini`](https://github.com/ihmeuw-msca/OneMod/blob/release/1.0/mypy.ini)?
- Do you want to keep [`noxfile.py`](https://github.com/ihmeuw-msca/OneMod/blob/release/1.0/noxfile.py) or parts of it at this time? I have not been using and I think we can just keep using `ruff` instead of `flake8` and `black`.

kels271828:
Fine to move mypy.ini into pyproject.toml; I agree that it's nicer to have everything in one place. I'm not sure about the missing imports, though. I don't think anyone on our team uses nox, so feel free to get rid of noxfile.py.

kels271828 (approved):

blsmxiu47:
OK thanks. Removing noxfile and the dependency-level mypy overrides for now as I haven't noticed it causing trouble not to have in the redesign and by default I would prefer not ignoring. Can add back in on a case-by-case basis if we deem it necessary going forward.

PR #110, opened 2024-10-22
Serialization module refactor

blsmxiu47:
## Description

Improvements to serialization functions logic, cleanliness. Adds unit tests for serializing, deserializing JSON and YAML.

## Changes made

- Rename `serializers` module to `serialiazation`
- Refactor `serialiazation` functions
- Add unit tests for `serialiazation` fns

## Additional notes / Open questions

- It was brought up that we might remove the `SerializableModel` base model if there is no point in having "preset" serialization methods for `Data`, `Constraint`, `ColumnSpec`, and any other Pydantic models. I am not yet sure whether there is a point to having these methods, but I haven't yet been able to convince myself that having `SerializableModel` does more harm than good. It's possible I am just forgetting the original justification for removing it, so remind me if there was one. It would not be hard to remove, but would require some adjustments to testing code, as there are OneMod `types` (e.g. `Data`) which currently use `from_json()`, etc. as convenience methods for serialization.


kels271828:
Some thoughts/questions on `SerializableModel`
- Since `Pipeline` and `Stage` need to write their own `to_json`/`from_json` methods anyway, I don't think they need to inherit from `SerializableModel`
- What's the use case for `Data`, `Constraint`, and `ColumnSpec` needing `to_json`/`from_json` methods? I think the only serialization/deserialization that happens right now is the pipeline config file, but correct me if I'm wrong.
- My vote for getting rid of `SerializableModel` is that it's unnecessary since we can use the `serialize`/`deserialize` methods, but I maybe I'm unaware of what classes inherit from it and how they use it

kels271828:
If `serialize` is only used to save a pipeline config file, I think `_normalize` and `_CustomJSONEncoder` may be unnecessary (see [dump_pipeline](https://github.com/ihmeuw-msca/OneMod/blob/ec21cc489e1468007ffac4af8a219066f1c3e8ce/src/onemod/serializers/functions.py#L102); I added this because `serialize` wasn't saving all of the stage config fields).

blsmxiu47:
Thanks for confirming. Removed `SerializableModel`. I don't think we have explicit use case at this time.

Do you want me to try integrating and testing `dump_pipeline()` as a part of this PR then or just be aware of it as a potential update coming from the `feature/implement_model_stages` branch? My initial thought however is that something `Pipeline`-specific might belong on `Pipeline` class.

The way I have written the serialization module so far is as a more "general-purpose" module, for instance by providing a single helper function for serializing both Pydantic models and Python dictionaries. In my branch `serialize()` is used to serialize non-Pydantic objects within the `Pipeline.build()` stage for saving 1) a validation report and 2) the Pipeline representation. I wouldn't mind replacing the saving of Pipeline representation bit with its own method (e.g. "dump_pipeline()" or similar), but I think in my mind that would still be on the `Pipeline` class, similar to `Pipeline.to_json()` is. Let me know what you think.

kels271828:
You can ignore `dump_pipeline` for now, I can move it to `Pipeline` on my branch!

PR #111, opened 2024-10-23
Implement Rover and SpXMod Stages

kels271828:
### Description

Implemented `RoverStage` and `SpXModStage` and simplified input/output load/dump.

### Changes Made

- local_backend.py: Moved local execution from `Pipeline` to its own module
- jobmon_backend.py: Added option for task-specific resources, fixed issues with upstream task lookup
- stage/base.py: Added `collect_after` property to `ModelStage`, replaced `directory` and `pipeline` properties with `dataif`, removed `to_json` method and `from_pipeline` args (now I'm always assuming a config file belongs to a pipeline rather than a specific stage), added separate `evaluate` method for `ModelStage`, added `get_field` method to grab items from pipeline or other stage configs, list groupby/crossby in `ModelStage` `repr` method
- config/base.py: Changed how stages inherit pipeline config items to simplify JSON output
- config/model_config/spxmod_config.py: Changed some items from set to list because of hash issues
- main.py: Added `init` function
- pipeline.py: Removed `groupby` field serializer, changed load/dump methods, moved subset and parameter set creation from `add_stage` to `build`, 
- utils: Added `metric.py`, `residuals.py`, and `uncertainty.py` from main branch
- subsets.py, parameters.py: Changed args from paths to data frames

### Notes

- Did not implement `KregStage`
- Did not implement optional `priors.pkl` input for `SpXModStage`
- Commented out validation method in `RoverConfig` because it needs to happen in `Pipeline.build`
- New modules in utils were previously in a "modeling" directory, should we do that again? Not sure if it's more cluttered to have lots of modules in utils vs. lots of different subpackages
- Previous version of OneMod used cache decorator for function that loads config file. I tried using that here (with `deserialize`), but ran into issues with the whole dictionary not loading and kept having to clear cache. Would be nice to figure this out, since we load the config file in multiple places


kels271828:
Not sure if I broke tests or just wasn't running them correctly; got a couple of `ModuleNotFoundError: No module named 'tests'` errors

blsmxiu47:
Not receiving the `ModuleNotFoundError`. It presumably stems from the fact that I added a `tests/utils` directory for shared helpers and some test scripts import from there, so if your `tests` directory is not recognized as a module you would run into this error on import. Could be related to `PYTHONPATH` or virtual environment, but I didn't do anything myself to set this up, other than make sure I had a fresh pip install using our `pyproject.toml` (`pip install -e .[dev]`, if the `dev` dependencies have been updated to account for your additions with this branch). Could be related to IDE too perhaps..

blsmxiu47:
The tests failures I do receive are of just two kinds:
1. `TypeError: Can't instantiate abstract class DummyStage without an implementation for abstract method 'run'`: I can probably fix these. I imagine I just need to update my dummy test stages to follow any updates you've made to the Stage classes.
2. `KeyError: ''` coming from `self.dataif.dump_output("parameters.csv", index=False)` (`pplkit` DataInterface) in `create_stage_params()` [here](https://github.com/ihmeuw-msca/OneMod/blob/feature/implement_model_stages/src/onemod/stage/base.py#L405). Will have to look into these ones. It's possible that I just need to update the test cases' syntax to match your updates.

kels271828:
I had more errors than the two you mentioned, but I updated tests based on syntax changes. There were a couple of places I commented things out (added fixme and description) where more than a quick fix was needed to pass.

blsmxiu47 (approved):

PR #112, opened 2024-10-29
Feat/GitHub ci workflow

blsmxiu47:
## Description

Adds GitHub Actions workflow with the following jobs:

- `lint`: Performs linting with `ruff` (no automatic fixes) on `src/`, `tests/`. Python>=3.10<3.13 (for now)
- `mypy`: Performs type checking with `mypy` (no automatic fixes) on `src/`, `tests/`. Python>=3.10<3.13 (for now)
- `test`: Runs all tests in `tests/` which do not have the mark `requires_data`. Python>=3.10<3.13 (for now)
- `build`: Builds OneMod with `build`. Python3.12 (for now)

Successful test run of this workflow: https://github.com/ihmeuw-msca/OneMod/actions/runs/11581822144

## Changes made

- Adds `.github/workflows/ci.yaml`
- Makes linting and `mypy` fixes for all redesign Python code under `src/` and `tests/`
- Updates any related tests affected by these `mypy` fixes
- Adds `requires_data` pytest marker for tests which need data assets stored outside of the repo

## Additional notes

- Some redesign implementation is still in-progress, so there are a couple places where `# type: ignore[...]` was added instead of an overhaul of the code to appease `mypy`, given that the code will change shortly and/or I might want to check in with Kelsey before updating certain functionality.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,139 @@
...
+            - "main"
+            - "release/*"
+    pull_request:
+        branches:
+            - "main"
+            - "release/*"
+    release:
+        types: [created]
+
+jobs:
```
zhengp0, line 15: Just curious, are there reasons that we want to keep these as separate jobs rather than steps?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,139 @@
...
+            - name: Run Tests
+              run: pytest --cov=src --cov-report=xml tests/ -m "not requires_data"
+
+            - name: Upload Coverage Report
+              uses: actions/upload-artifact@v4
+              with:
+                name: coverage-report-${{ matrix.python-version }}
+                path: coverage.xml
+
+    build:
```
zhengp0, line 109: Do we need trigger `build` every time when push to `main`?

src/onemod/constraints/functions.py
```diff
@@ -3,79 +3,94 @@
 
 from polars import Series
 
-def bounds(ge: float = None, le: float = None) -> Callable[[Series], None]:
+
+def bounds(
+    ge: float | None = None, le: float | None = None
```
zhengp0, line 8: Really minor point, I think it will be great we use `lb` and `ub` compare to `ge` and `le`, more widely used in our context.

src/onemod/dtypes/column_spec.py
```diff
@@ -1,18 +1,39 @@
-from typing import List, Type
+from typing import Any, List, Type
```
zhengp0, line 1: Minor point here as well, we use `List` and `list` for typing, also `Dict` and `dict`. I think we should all switch to `list` and `dict`.

src/onemod/dtypes/column_spec.py
```diff
@@ -1,18 +1,39 @@
...
+        elif key == "constraints":
+            return self.constraints
+        else:
+            raise KeyError(f"Invalid key: {key}")
+
+    def __contains__(self, key: str) -> bool:
+        return key in {"type", "constraints"}
+
+    def keys(self):
+        return ["type", "constraints"]
```
zhengp0, line 31: I might be really out of this, but curious to know why we want to make `ColumnSpec` dictionary like?

src/onemod/serialization/functions.py
```diff
@@ -1,11 +1,13 @@
 import json
 from pathlib import Path
-from typing import Any, List
+from typing import Any, TypeVar
 
 import yaml
 from pydantic import BaseModel
 from yaml import Node, SafeDumper
 
+T = TypeVar("T", bound=BaseModel)
```
zhengp0, line 9: Really cool, never use a generic type before :)

zhengp0 (approved):
Thanks very much @blsmxiu47! This looks good, I left some minor comments. A random thought point that might not be for this PR, when we use `Literal[...]` is it better to use an enum and maybe pattern matching like in rust? It will at least automatically error out when the input is not in the set.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,139 @@
...
+            - name: Set up Python
+              uses: actions/setup-python@v5
+              with:
+                python-version: ${{ matrix.python-version }}
+
+            - name: Install dependencies
+              run: pip install .[dev]
+
+            - name: Run Mypy
+              run: mypy src/ tests/
```
zhengp0, line 72: We also have `ruff` and `mypy` in the pre-commit, it is better to catch these errors before we push the repo I think.

.github/workflows/ci.yml
blsmxiu47: A few reasons I think (though maybe there are reasons to flatten the workflow into steps that I'm not aware of, please lmk!):

- For non-dependent tasks jobs can run in parallel, which can make the workflow run faster. Also for non-dependent tasks, jobs allow us to provide a clean environment for running each, which improves reproducability and could make it easier to understand whats going wrong in some failure scenarios.
- In the Actions failure reports, individual jobs' statuses will reported at the top level, which personally I like to be able to view at a glance. I like to be able to differentiate whether linting vs. type checking failed, for instance, without clicking in.
- Each job has its own log output, which, maybe this is a preference, but I think is nicer than one long dump of the workflow logs, especially as you add more tasks.
- Though we am not making use of this (yet), we could run specifically jobs conditionally. For example only running `build` on releases, as I think you might be suggesting below (and I think I agree).

.github/workflows/ci.yml
blsmxiu47: No I agree we should probably only trigger build on release. There are two ways we could accomplish this - either 1) add a conditional line to the same file so that `build` job only triggers when the event is release, or 2) we can separate `build` into a new workflow/config file (e.g. `build.yml`). I kinda like the latter because it makes it obvious that we have two sets of jobs which run in different situations, but not opinionated

src/onemod/constraints/functions.py
blsmxiu47: Good to know! I have not thought about the constraint functions we have at all, just made them for the sake of testing core validation functionality really, so I would like to follow up with you to pick your brain on what useful constraint functionality we could add to OneMod out of the box.

src/onemod/dtypes/column_spec.py
blsmxiu47: Agree with the consistency approach. I did not pay attention to that on this branch but actually it does make decent sense to go through replace instances of this sort of inconsistency here. brb.

src/onemod/dtypes/column_spec.py
blsmxiu47: There is a longer story to this where `ColumnSpec` started out as a `TypedDict` way back when.. But now the only reason really is because we are accessing it like a dictionary elsewhere in the package.

.github/workflows/ci.yml
blsmxiu47: Yes I have set up both intentionally, because there is technically nothing stopping someone from explicitly ignoring pre-commit when making their commits. But agree we should not be doing that as a rule of thumb 😄

blsmxiu47:
@zhengp0 RE: `Literal[...]` vs. adding enums - I do like this idea. I think most of the implementation using `Literal` in type hints was added as a means to get to a working MVP quickly. To be fair, we could add runtime validation even without using enums, but enums could make a lot of sense especially because I think the same specifications are reused multiple times in some cases (e.g. `backend: Literal["local", "jobmon"]`, `method: Literal["run", "fit", "predict", "collect"]`). Adds a bit of overhead but feels natural in this context to me (but agree not in this PR).

PR #113, opened 2024-11-01
Evaluating a subset of OneMod Pipeline Stages

blsmxiu47:
## Description

Allows users to run `Pipeline.evaluate()` with a `stages` argument to specify a subset of the Pipeline's Stages to be evaluated. Per design requirements, will raise an error when stages are specified but any required inputs (e.g. upstream dependency output(s)) for the Stage do not exist.

Decided to maintain the passing of `model` (i.e. a `Pipeline` or `Stage`) to `evaluate_local()` and `evaluate_with_jobmon()`, for 1) simplicity early on and 2) to avoid repeating building/validation unnecessarily. Alternatively we could implement the concept of a `Dag` object specifically for an intermediate to execution and distinct from `Pipeline`, which may or may not be a full `Pipeline`, however at this time it seems that the use case for running subsets of a `Pipeline` is more a feature to allow testing of subsets + resuming evaluation of a partially-run `Pipeline`, rather than demand for constructing and validating "sub-Pipelines" that would represent complete modeling pipelines and essentially be shareable as new Pipelines in their own right. If there is demand for this, do let me know, but not something I've heard yet and seems superfluous for known use cases.

## Changes made

- Adds `stages` arguments to `Pipeline.evaluate()`
- Adds related `stages` arguments and handling to `evaluate_local()` and `evaluate_with_jobmon()`
- Adds integration tests for evaluating subsets of stages of a OneMod `Pipeline`
- Improves organization of test dirs/scripts
- Minor updates to `pipeline_example.py` to clarify unlocked functionality
- Adds `requires_jobmon` pytest marker

## Additional notes / considerations

- Adds a TODO to be addressed in the "Migrating `DataInterface` to OneMod" ticket, as determining existence of stage input data files probably will likely make more sense living in this future module, rather than entirely handled within the `Stage` base class.

### DIFF
--- a/src/onemod/pipeline.py
+++ b/src/onemod/pipeline.py
@@ -136,6 +136,19 @@ class Pipeline(BaseModel):
         stage.config.inherit(self.config)
         self._stages[stage.name] = stage
 
+    def check_upstream_outputs_exist(
+        self, stage_name: str, upstream_name: str
+    ) -> bool:
+        """Check if outputs from the specified upstream dependency exist for the inputs of a given stage."""
+        stage = self.stages[stage_name]
+
+        for input_name, input_data in stage.input.items.items():
+            if input_data.stage == upstream_name:
+                upstream_output_path = input_data.path
+                if not upstream_output_path.exists():
+                    return False
+        return True
+
     def get_execution_order(self) -> list[str]:
         """
         Return topologically sorted order of stages, ensuring no cycles.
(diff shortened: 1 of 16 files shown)


src/onemod/pipeline.py
```diff
@@ -281,20 +286,42 @@ def evaluate(
...
+                    raise ValueError(
+                        f"Stage '{stage_name}' not found in pipeline."
+                    )
+
+            for stage_name in stages:
+                stage: Stage = self.stages.get(stage_name)
+                if not stage.check_required_inputs_exist():
+                    raise ValueError(
+                        f"Required inputs for stage '{stage_name}' are missing. Ensure data dependencies are available before execution."
+                    )
```
zhengp0, line 307: For the logic here, if we have stage `A`'s inputs depend on stage B's output and the stage `B` is in the subset, we might not require the input from stage `A`. To make the example more concrete, assume we have stage, `A`, `B`, `C` and we have our pipeline as
```
C -> B -> A
```
If we our subset is `{A, B}` we will only require input for `B`. Please let me know if I mis-understood anything

zhengp0 (commented):
Thanks very much @blsmxiu47 ! This PR looks pretty good, I only have a minor question on checking the input. Please let me know what you think.

src/onemod/pipeline.py
blsmxiu47: You are correct, thanks for catching! I had written a test that should only pass if this was the case but then sabotaged it with a line manually adding the input data prior to `evaluate()`..

Anyway, updated the validation and the test case. There probably is a more efficient/sophisticated way to run this check on upstream dependencies data outputs, but I think it might be a good role for `DataInterface` so haven't spend time on thinking of optimizing the process here. But for now it is a method on `Pipeline`. Let me know if you think this is sufficient for the time being, or if you think it deserves more iteration sooner rather than later.

zhengp0 (approved):
Thanks very much @blsmxiu47! It looks good

PR #114, opened 2024-11-12
DataInterface, id_subsets, and Polars Support

blsmxiu47:
## Description

Allows users to specify `id_subsets` in calls to `Pipeline.evaluate()` (and also `Pipeline.build()`). `id_subsets` is passed to `create_stage_subsets()` for each model stage in the Pipeline, so that stage-level subsets creation (and saving of `subsets.csv` for instance) is affected by this argument.

Now that we are getting into loading and running stages for subsets of data, I have used this opportunity as an initial POC for trying out Polars, and specifically use of the LazyAPI internally within OneMod. I have migrated functionality from `pplkit.data` into a new OneMod module called `fsutils` (I don't love this module name, but didn't want to steal "pplkit" and also didn't want to call the module "data". Also we already have a module called "io". Very open to alternative naming ideas, but what the module basically is is a paths manager and manager of filesystem read/write).

Along the way I made polars the default backend for `DataInterface`, but included support for pandas. I have used polars LazyFrames for the `id_subsets` implementation as a proof of concept, but it would be trivial to switch back to pandas at this early stage if it proves more accessible.

## Changes made

- Adds `id_subsets` argument to `Pipeline.evaluate()`, `Pipeline.build()`, and `Stage.create_stage_subsets()`.
- Adds a new module - `fsutils` (again, could call it something else) - to replace use of `pplkit.data` for the `DataInterface`.
- Separates the concerns of `DataLoader` and `ConfigLoader` internally, though preserves the `stage.dataif.load(...)` syntax with a "universal" `DataInterface`, for consistency of use.
- Removes data interface dynamic attributes in favor of a simple dictionary approach (see `PathManager`), due to concerns over clarity, error traceability, and documentability of dynamic attributes.
- Adds additional functionality to `DataInterface`, especially wrt data loading, such as support for `columns` arg, `id_subsets` arg, `lazy` option for lazy loading with Polars, and `backend` arg (polars or pandas).
- Adds related tests, though probably haven't exhausted edge cases; want to get POC review first and design-focused feedback before going crazy on the testing.

## Additional notes

- The majority of review should probably focus on the new module which contains the OneMod implementation of `DataInterface`. I have made some opinion-based decisions, for example removing dynamic attributes from `DataInterface` and related classes, as well as separating the concerns of `DataLoader` vs. `ConfigLoader` internally.
- `DataLoader` and `ConfigLoader` each have distinct logic required, and for instance data object loading now requires more arguments than the similarly named methods for config files. I could also see us adding more distinct options for each down the road, so I think this proposed implementation can improve scalability. As a compromise, I preserved `DataInterface` as a unified interface for data+config assets, even though I personally find the name a bit confusing.


src/onemod/fsutils/config_loader.py
```diff
@@ -0,0 +1,28 @@
...
+    def dump(self, obj: Any, path: Path, **options) -> None:
+        """Save a config or model object."""
+        if path.suffix not in self.io_dict:
+            raise ValueError(f"Unsupported config format for '{path.suffix}'")
+
+        self.io_dict[path.suffix].dump(obj, path, **options)
+
+    def __repr__(self) -> str:
+        base_repr = super().__repr__()
+        return f"{base_repr[:-1]}, handles_configs=True\n)"
```
zhengp0, line 28: Since `ConfigLoader` is not subclass for anything, `base_repr` will be something like `<...>` I don't know if it will play well with the additional `handles_configs` appended at the end.

src/onemod/fsutils/data_loader.py
```diff
@@ -0,0 +1,77 @@
...
+    """Handles loading and dumping of data files with optional lazy loading,
+    column selection, and subset filtering."""
+
+    io_dict: dict[str, DataIO] = dataio_dict
+
+    def load(
+        self,
+        path: Path,
+        lazy: bool = False,
+        backend: Literal["polars", "pandas"] = "polars",
```
zhengp0, line 20: One random idea is that, we can combine arguments `lazy` and `backend` to
```python
backend: Literal["polars", "polars_lazy", "polars_eager", "pandas"] = "polars"
```
where `"polars"` is an alias as `"polars_lazy"`, this way we can reduce a argument, and don't have to worry about user input `pandas` and the backend and `lazy=True`.

src/onemod/fsutils/data_loader.py
```diff
@@ -0,0 +1,77 @@
...
+    ) -> pl.DataFrame | pl.LazyFrame | pd.DataFrame:
+        """Load data with optional lazy loading and subset filtering. Supports
+        both Polars and Pandas backends."""
+        if lazy and backend == "pandas":
+            raise ValueError("Pandas backend does not support lazy loading")
+
+        if path.suffix not in self.io_dict:
+            raise ValueError(f"Unsupported data format for '{path.suffix}'")
+
+        if backend == "polars":
```
zhengp0, line 33: Here we can always use the lazy frame to load and if `"polars"` eager we `collect` and if `"pandas"` we collect and convert to pandas data frame :)

src/onemod/fsutils/interface.py
```diff
@@ -0,0 +1,70 @@
...
+from typing import Any, Literal
+
+import polars as pl
+
+from onemod.fsutils.config_loader import ConfigLoader
+from onemod.fsutils.data_loader import DataLoader
+from onemod.fsutils.path_manager import PathManager
+
+
+class DataInterface(PathManager):
```
zhengp0, line 11: I am definitely ok to change this to `FileInterface` to reduce the probability for confusion.

src/onemod/fsutils/io.py
```diff
@@ -0,0 +1,227 @@
...
+
+import dill
+import pandas as pd
+import polars as pl
+import tomli
+import tomli_w
+import yaml
+
+
+class DataIO(ABC):
```
zhengp0, line 14: As we discussed, one thing we can consider is to use `FileIO` as the name, and only provide `csv, parquet` and other potential tabular data format with different backend options :)

src/onemod/fsutils/path_manager.py
```diff
@@ -0,0 +1,41 @@
+from pathlib import Path
+
+
+class PathManager:
```
zhengp0, line 4: Several random thoughts here :)
* This class is very similar to the dictionary class, maybe we can directly subclass `dict` to borrow some existing methods.
* One thing I found I frequently doing is to build path on the existing path. For example if we specify the `root` folder, we can create `data` or `config` folder based on the root folder. Might be good to add a method for it

zhengp0 (commented):
Hey @blsmxiu47! Thanks very much for the PR! I put in some minor comments, please let me know what you think :)

src/onemod/fsutils/config_loader.py
blsmxiu47: Ah, yup I did not update `__repr__()` after making certain design changes thanks 👍

src/onemod/fsutils/data_loader.py
blsmxiu47: This is an interesting proposition. It sounds like something I would want to do in order to simplify code and make maintaining this logic easier, esp if we end up expanding `DataInterface` and `load()` responsibilities in the future, which seems likely. So there are certainly potential pros. In terms of memory and performance, I expect it would usually be a net improvement, but I actually bet it depends. There is non-zero overhead in a to_pandas conversion step.

More importantly, I think this would be fundamentally changing what is meant by "backend", to the point that we might even call it something else. Admittedly I have already pretty much done this with `"polars_lazy"` vs `"polars_eager"`.. so the way I have done that is actually questionable in the opposite direction I think.

But what I mean is I think this proposition would not come without tradeoffs. I added the `pandas` logic separately because I imagine if someone (let's say they are unfamiliar with polars) specified `"pandas"` for `backend` and they ran into loading errors (whether at reading, subsetting, whichever stage), they would expect the errors they receive to be the Pandas errors they are familiar with, right? Whereas if behind the scenes we are actually still using LazyFrames, the issues they might run into could be very confusing for them (like "I explicitly chose Pandas backend, so why am I getting Polars errors?"). 
Now, we could change the argument name to "return_type" or something, to make it less misleading, but then there's no option for pandas backend at all. This is not something that I personally would care about, since I don't think I would personally use pandas for most big data tasks and I am comfortable receiving polars errors and warnings from my load function, so I will defer to you and others on what seems the most useful and intuitive from the average user's perspective

src/onemod/fsutils/interface.py
blsmxiu47: If this is in response to my naming-related mutterings, I really don't mind too much ^^; I still like the ring of `DataInterface`, and frankly `dataif`, better, so if you don't mind I actually prefer not to change this for now.

However happy to change it if you feel more strongly about this.

src/onemod/fsutils/io.py
blsmxiu47: Yes I think this is a good idea. I discovered some additional complexity though I think, so do you mind if I make a new ticket for this? Don't think I'll be able to quite get to it this week.

src/onemod/fsutils/path_manager.py
blsmxiu47: I do enjoy your random thoughts :D

- Sure. What methods/use cases are you specifically thinking of?
- Could you elaborate on this? I think I haven't fully understood the idea. How would it be different from using `pathlib` for instance to construct paths?
  - Maybe related - is there any value in having a special `root` directory? I guess this is kind of an idea on an idea. But on second thought as long as our directories management system is intuitive enough it probably doesn't add value.

src/onemod/fsutils/data_loader.py
zhengp0: Thanks very much for being so thoughtful @blsmxiu47! And it is a good point, indeed if we change the behavior to always use `polar_lazy`, `backend="pandas"` might contradict with what user expect. I like the idea of changing it to `return_type` and always sticking with `backend=polar_lazy`.

src/onemod/fsutils/interface.py
zhengp0: That sounds good!

src/onemod/fsutils/io.py
zhengp0: Yes! That sounds good! This is not critical at all, just nerding out on the naming :)

src/onemod/fsutils/path_manager.py
zhengp0: For the dictionary inherent idea. It is more for simplifying the code by subclassing `collections.UserDict`. We only need to create a `add` method, and overwrite `__setitem__` method, and for the `get_path`, `remove_path` we will have the counterpart from dictionary `__getitem__` or `pop` and we can batch update with the `update` as well. I don't have a strong preference on this, just a hutch it might be more future proof.

For the path dependency idea, an concrete example is that we currently have a parent folder called `directory`, and all the `data`, `config` directory is provided relative to the `directory` folder. This allows user to repetitively specify the full path from `directory` and build the path from the existing paths in the object. But as you said, we can just easily explicitly use the path in the object `path_manager.get_path("directory") / "data"` to achieve this as well, and probably more readable and flexible this way. Let me know what you think :)

zhengp0 (approved):
Thanks very much for addressing my comments @blsmxiu47!

src/onemod/fsutils/path_manager.py
blsmxiu47: Would the alternative be like an `dataif.root.append_path("data")`? I think it works if there is a special concept of `root`, and maybe there is other future functionality to add around it. Sounds like this is what you are implying, correct? I suppose then we have methods for `get_root()`, `set_root()`.
Or else is it that with `add_path()` we assume all paths are relative to `root` by default and the user needs to specify if they want to provide a non-relative path? This is one of the first implementation approaches I went with. But then I realized there was at least one place already in OneMod where an absolute path (pointing to the Pipeline data file) was specified and after trying out each approach it felt more intuitive (to me, with current OneMod Pipeline, Stages relationship) to assume absolute path by default.
For arbitrary non-special directories, just using `get_path()` probably makes sense.

I'm going to propose another ticket on this one too ^^; Might be a small amount of design discussion still to be had.

src/onemod/fsutils/data_loader.py
blsmxiu47: Updated to this approach. Code does look cleaner now, thanks! But let us know if there is ever a demand for pandas (or other) backend.

Left the argument as a string literal. Not sure if it would be better practice to actually require a Type to be specified for `return_type` (e.g. `pl.DataFrame` instead of `"polars_dataframe"`). Either way works, but the non-string method requires the library to be imported for instance (which it probably would be, but still..) Anyway want to do whatever is easiest and most intuitive from the users' POV. lmk if you want me to change it.

src/onemod/fsutils/path_manager.py
zhengp0: That sounds good! Thanks very much @blsmxiu47!

src/onemod/fsutils/data_loader.py
zhengp0: This looks good! Thanks very much!

src/onemod/backend/jobmon_backend.py
```diff
@@ -231,7 +231,7 @@ def evaluate_with_jobmon(
 
     # Set config
```
kels271828, line 232: I think the config arg for jobmon should be a path, not a config object (hence the previous `model.dataif.config`)

kels271828: They should both be the path to the pipeline config file. Where does it expect a `PipelineConfig` object?

blsmxiu47: Just below, at https://github.com/ihmeuw-msca/OneMod/blob/release/1.0/src/onemod/backend/jobmon_backend.py#L236.

Because if `model` is a `Pipeline`, then `model.config` would be a `PipelineConfig` - https://github.com/ihmeuw-msca/OneMod/blob/release/1.0/src/onemod/pipeline.py#L42. This attribute on `Pipeline` has been the case since the initial redesign commits I think.

blsmxiu47:
Merging this PR as it has gotten big. But noting that there is at least one unresolved comment and there will need to be fixes, esp. jobmon-related. Also a few new tickets to be made based on conversations on this PR. So, basically, more to come!

src/onemod/backend/jobmon_backend.py
kels271828: - `config` is an argument for `evaluate`, and should be the path (string) to the pipeline config JSON file: https://github.com/ihmeuw-msca/OneMod/blob/ec384608a9bbe8815112c76749b717111fbec470/src/onemod/main.py#L124
- Here's the command line argument that Jobmon uses to call `evaluate`: https://github.com/ihmeuw-msca/OneMod/blob/ec384608a9bbe8815112c76749b717111fbec470/src/onemod/backend/jobmon_backend.py#L143
- Here's what `task_args` used to look like, where `config` is a string: https://github.com/ihmeuw-msca/OneMod/blob/1a91608e2faede8adc43c5c5de746e9f4907c834/src/onemod/backend/jobmon_backend.py#L223

I think it might be confusing that I used "config" to mean both the JSON file and the config objects for `Pipeline` and `Stage`; maybe we should use `config_path` or something different to make it more obvious?

PR #116, opened 2024-12-06
Some fixes to make codem pipeline work

zhengp0:
## Description

- Fix minor typo
- In `IO.check_exists` add logic branch when `item_value.stage not in upstream_stages`
- In `_get_stage`, make sure that customized stage have higher priority compare to built-in stages

## Discussion

If user define a customized stage with the same with as one of the built-in stage, `module` property won't get created, and it is causing trouble with the CODEm pipeline `PreprocessinStage`. I suggest we always save the module path and load stage with module path. Let me know what you guys think


kels271828:
> I suggest we always save the module path and load stage with module path. Let me know what you guys think

It seems unnecessary to save the module path if it just points back to the OneMod package, that's just my opinion.

I removed the preprocessing stage since I don't think it makes sense to have a "general" one in the package anyway.

kels271828 (approved):
Looks good, thanks for fixing these issues!

PR #117, opened 2024-12-10
Input existence checking

zhengp0:
## Description

Stop checking input existence if the input is part of the output from upstream stage that hasn't been produced yet.

Please let me know if I misunderstood something here.

### DIFF
--- a/src/onemod/io/base.py
+++ b/src/onemod/io/base.py
@@ -164,22 +164,20 @@ class Input(IO):
         FIXME: Assumes pipeline has been built (so paths are absolute)
 
         """
-        item_names = item_names or set(self.items.keys())
-        upstream_stages = upstream_stages or self.dependencies
+        if item_names is None:
+            item_names = set(self.items.keys())
+        if upstream_stages is None:
+            upstream_stages = self.dependencies
 
         missing_items = {}
         for item_name in item_names:
-            item_value = self.__getitem__(item_name)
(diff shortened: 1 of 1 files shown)


kels271828:
Were you getting errors?

The way I wrote it, the arg `upstream_stages` are the stages that you DO want to check (see description in docstring: https://github.com/ihmeuw-msca/OneMod/blob/aa648a90ee21f85a449cba1f982f306b53f7aaa9/src/onemod/io/base.py#L155)

When the function gets called in `Pipeline.evaluate()`, it only passes the stages that aren't being run (https://github.com/ihmeuw-msca/OneMod/blob/aa648a90ee21f85a449cba1f982f306b53f7aaa9/src/onemod/pipeline.py#L346).

I also added `check_exists()` to `Stage.evaluate()` (https://github.com/ihmeuw-msca/OneMod/blob/aa648a90ee21f85a449cba1f982f306b53f7aaa9/src/onemod/stage/base.py#L262) and `ModelStage.evaluate()` (https://github.com/ihmeuw-msca/OneMod/blob/aa648a90ee21f85a449cba1f982f306b53f7aaa9/src/onemod/stage/base.py#L561). This is redundant if you've called `Pipeline.evaluate()`, so we can check the logic of when we check stage input. 

The main reasons to have the `check_exists()` in the first place is to 1) have more informative error messages, and 2) to fail before running or scheduling anything.

kels271828 (approved):

PR #118, opened 2024-12-11
feature/jobmon_resources

kels271828:
Allow user to pass a dictionary of compute resources or a path to a resources file (any format that can be loaded by `ConfigLoader`) to `evaluate_with_jobmon` and pipeline/stage `evaluate` methods. Previously only paths to yaml files were valid.

### DIFF
--- a/src/onemod/backend/jobmon_backend.py
+++ b/src/onemod/backend/jobmon_backend.py
@@ -34,7 +34,6 @@ import sys
 from pathlib import Path
 from typing import Literal
 
-import yaml
 from jobmon.client.api import Tool
 from jobmon.client.task import Task
 from jobmon.client.task_template import TaskTemplate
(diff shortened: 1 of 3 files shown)


zhengp0 (approved):
Thanks @kels271828! This looks good to me!

src/onemod/backend/jobmon_backend.py
```diff
@@ -34,7 +34,6 @@
 from pathlib import Path
 from typing import Literal
 
-import yaml
```
zhengp0, line 36: Highly recommend ruff pre-commit
https://github.com/astral-sh/ruff-pre-commit
it will automatically detect the unused import and fix it :)

kels271828: I had formatting on save but it wasn't working today for some reason. I had to update a vscode extension and now it is working again :)

PR #119, opened 2024-12-11
"Fix" mypy issue in CI

zhengp0:
## Descriptions

- Skip `mypy` checking for `utils.residual`, `utils.uncertainty` and `model_stages.rover_stage` due to Pandas typing issue
- Add `None` type to `constraints.base.Constraint.func` because default value is `None`

### DIFF
--- a/src/onemod/constraints/base.py
+++ b/src/onemod/constraints/base.py
@@ -13,7 +13,7 @@ class Constraint(BaseModel):
     name: str
     args: dict[str, Any]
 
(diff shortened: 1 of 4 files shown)


kels271828:
We should create a ticket to fix the mypy issues with pandas

kels271828 (approved):

zhengp0:
> We should create a ticket to fix the mypy issues with pandas

Yes! Will do!

PR #120, opened 2024-12-16
Fix pre-commit mypy and standardizing environment setup

blsmxiu47:
## Description

@zhengp0 mentioned issues with local environment setup (I think conflicting `conda` envs iirc) and difficulty pointing to the correct `mypy` executable while developing. I also noticed that `mypy` would need to be run manually (which is not bad practice, but still) prior to commit in order to catch typing issues early, rather than it automatically.

So,
1) fixed an incorrect line in `.pre-commit-config.yaml` to ensure it always runs for all python files in `src/` and `tests/` (files updated within the commit in question, that is) to enable automatic pre-commit type checking.
2) More significantly, added a `Makefile` that provides some environment setup-related targets that may be useful in order to ensure the local dev environment is set up correctly. As a part of this PR I removed my previous conda environment and tested setting up in this manner (i.e. `make setup ENV_TYPE=conda`, then `conda activate onemod`) and can confirm that my `mypy`, `pytest`, etc. executables all point to the expected locations within the new (`"onemod"`) conda environment so please do let me know if this approach still does not work for you using this branch.

## Changes made

- One-line regex fix to `.pre-commit-config.yaml`
- Adds `Makefile` with (optional but maybe useful) environment setup targets
- Adds drafts for some getting started dev documentation for the `Makefile` additions, though have not yet approached the task of rewriting all dev docs for the redesign
- Adds a potential `CONTRIBUTING.md` skeleton for discussion. Would want to discuss contribution guidelines together though before filling in. May not be highest priority atm.


Makefile
```diff
@@ -0,0 +1,59 @@
...
+			echo "Error: Python $(PYTHON_VERSION) is not installed."; \
+			exit 1; \
+		fi; \
+	fi
+
+setup: check-python  ## Set up the development environment (venv or conda)
+ifeq ($(ENV_TYPE),venv)
+	python$(PYTHON_VERSION) -m venv $(VENV_DIR)
+	$(VENV_DIR)/bin/pip install -U pip
+	$(VENV_DIR)/bin/pip install -e ".[dev]"
```
kels271828, line 26: Could we have a way to install other optional dependencies with the make file (e.g., jobmon, docs...)?

docs/developer_guide/setup.rst
```diff
@@ -0,0 +1,143 @@
...
+
+5. Start Developing!
+--------------------
+
+Congratulations! You’re ready to start contributing to `OneMod`.
+
+To manually run development tools, you can use the `Makefile` shortcuts:
+
+- **Run `mypy` for type checking**:
+  .. code-block:: bash
```
kels271828, line 116: When viewing on github, you can see ".. code-block:: bash"

.pre-commit-config.yaml
```diff
@@ -14,5 +14,5 @@ repos:
     rev: v1.12.0
     hooks:
       - id: mypy
-        files: ^src ^tests
+        files: ^(src|tests)
         args: ["--config-file", "pyproject.toml", "--install-types"]
```
kels271828, line 18: Remove "--install-types", add `additional_dependencies: [types-PyYAML]`

kels271828 (approved):

kels271828:
Thanks for setting all of this up!

kels271828:
Does anyone else get this error when running ruff:

warning: The isort option `isort.split-on-trailing-comma` is incompatible with the formatter `format.skip-magic-trailing-comma=true` option. To avoid unexpected behavior, we recommend either setting `isort.split-on-trailing-comma=false` or `format.skip-magic-trailing-comma=false`.

I can fix this by adding this to pyproject.toml:

[tool.ruff.lint.isort]
split-on-trailing-comma = false

Just wasn't sure if I'm the only one seeing this error.

kels271828:
After running pre-commit, I get a lot of files changed because of the end-of-file-fixer. Should we add those file changes to this PR?

blsmxiu47:
We can definitely remove the `make mypy`, `make pre-commit` targets before merging this branch; first just want to ensure that @zhengp0 can get setup successfully.

Makefile
```diff
@@ -0,0 +1,59 @@
+ENV_TYPE ?= venv
+PYTHON_VERSION ?= 3.10
+VENV_DIR := .venv
```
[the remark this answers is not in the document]
blsmxiu47: This comment might be too late now, but with `:=` the variable is immediately assigned the value at the time it is declared, and it cannot be overridden later when invoking make, whereas `?=` the value is only assigned if the var is not already defined (either in the Makefile itself or as a command-line override when invoking make).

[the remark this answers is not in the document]
kels271828: Ok, that's what I thought from looking it up, I was just confused because when I called `make list-args VENV_DIR=dummy` it prints the overridden value

Makefile
blsmxiu47: Yes. Not sure whether that's additional options we need to add to the Makefile. I don't mind it, but personally I'd probably just install manually with `pip install -e ".[docs]"`, etc. after environment setup

kels271828: Okay, it's fine to keep things simple

blsmxiu47:
@kels271828 Added in the 

```
[tool.ruff.lint.isort]
split-on-trailing-comma = false
```

as you suggested. I was also getting that warning.

Also fixed some of the docs syntax, added `.env` vars note to the dev docs, and as for the `:ref:`s, these links look and work fine when building and viewing locally, but not sure if there's a way to get them functioning when viewing in GitHub(?) Would be interested if you know of a way 😅

kels271828:
> Also fixed some of the docs syntax, added `.env` vars note to the dev docs, and as for the `:ref:`s, these links look and work fine when building and viewing locally, but not sure if there's a way to get them functioning when viewing in GitHub(?) Would be interested if you know of a way 😅

Maybe it's just an issue with viewing on GitHub; I was just viewing it in the browser not actually building the documentation with sphinx

zhengp0 (approved):
Thanks @blsmxiu47 ! Run through with no issue :)

PR #122, opened 2024-12-22
Fix/pandas mypy errors

blsmxiu47:
## Description

Addressing [MSCA-366](https://jira.ihme.washington.edu/browse/MSCA-366). For the most part came down to replacing instances of `df.eval("some string-based expression here")` with explicit pandas operations. In general, I would argue that we should avoid using strings to represent data operations, as it is more error-prone, harder to debug, and not static type-safe.

## Changes made

- Remove all instances of `# mypy: ignore-errors`
- Replace all instances of `pandas.DataFrame.eval()` in the onemod source code with explicit Python code
- In certain cases, broke operations out into a few lines for code readability and to ensure I was understanding the totality of the operation

## Additional notes

- Merged updates from #121 to make sure the tests updates and fixes from that branch are still valid with these changes. Preferably that one gets reviewed first

### DIFF
--- a/src/onemod/fsutils/data_loader.py
+++ b/src/onemod/fsutils/data_loader.py
@@ -17,33 +17,47 @@ class DataLoader:
         self,
         path: Path,
         return_type: Literal[
-            "polars_dataframe", "polars_lazyframe", "pandas_dataframe"
-        ] = "polars_dataframe",
+            "pandas_dataframe", "polars_dataframe", "polars_lazyframe"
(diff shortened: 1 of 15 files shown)


kels271828:
That's too bad that mypy doesn't like `DataFrame.eval()`; I think we started using that because it had better performance. But it makes sense that using strings would be more error-prone.

kels271828 (approved):

PR #124, opened 2024-12-27
Remove required fields from config classes

kels271828:
**Summary**
- Removed required fields from config classes for more flexibility
- `get()`, `__getitem__()`,`__contains__()`, and `__repr__()` methods ignore fields with None values
- `StageConfig` uses `StageConfig._pipeline_config` field values if stage field doesn't exist or is None

**Changes**
- Removed `PipelineConfig` and `ModelConfig` classes
- Removed required fields for `Config` and `StageConfig` (any field can be added to a config class during initialization or using the `__setitem__` method, there just won't be any field validation unless user writes their own subclasses)
- Removed unused config/data_config and stage/data_stages directories
- Changed `stage.config._pipeline` to `stage.config._pipeline_config`
- Changed `stage.config.inherit()` to `stage.config.add_pipeline_config()`
- Updated model config and model stage classes (Kreg, Rover, SPxMod), added checks for required fields in `StageConfig.add_pipeline_config()`
- Updated relevant tests and test helpers, added unit tests and integration tests
- Added `StageConfig.stage_contains()`, `StageConfig.pipeline_contains()`, `StageConfig.get_from_stage()`, and `StageConfig.get_from_pipeline()` methods
- Added `Config._get_fields()`, `StageConfig._get_fields()`, `StageConfig._get_stage_fields()`, and `StageConfig._get_pipeline_fields()` (methods exclude fields with None values)
- Added `__repr__` methods that exclude fields with None values

**Notes to reviewers**
- Previously, `PipelineConfig` had required fields that would get passed to `StageConfig`/`ModelConfig` instances if the stage config fields were None. Now that we removed all required fields from `Config` (which is what `Pipeline` classes use now that I removed the `PipelineConfig` class), I added a `_required` field to `StageConfig` that subclasses can fill in to make sure all required fields are present in either the stage or pipeline instances (check in `StageConfig.add_pipeline_config()`). We wouldn't need these checks if we removed the None defaults to the required model config fields, but then we wouldn't be able to inherit global settings from the pipeline's config (i.e., any global setting would have to be copy/pasted into all model stage instances, which isn't as clean and is more error-prone). Seems to work, but there might be a nicer way to do all of this.
- The stage- and pipeline-specific `StageConfig` methods may never be used, but felt like adding them for completeness. Can remove if you guys think it's cleaner.


src/onemod/config/model_config/rover_config.py
```diff
@@ -2,24 +2,52 @@
 
 FIXME: improve top_pct_score and top_pct_learner descriptions
 
+TODO: Update docstrings, default is None, if None, must inherit from
```
blsmxiu47, line 5: Does this docstring update require a separate ticket, or perhaps part of the documentation tasks?

src/onemod/config/base.py
```diff
@@ -1,156 +1,124 @@
...
 
     """
 
-    model_config = ConfigDict(extra="allow")
+    model_config = ConfigDict(
+        extra="allow", validate_assignment=True, protected_namespaces=()
+    )
+
+    _pipeline_config: Config = Config()
+    _required: set[str] = set()  # TODO: unique list
```
blsmxiu47, line 63: Noted - will be sure to address these when I ultimately merge these changes into my UniqueList branch

src/onemod/config/model_config/rover_config.py
```diff
@@ -41,6 +69,12 @@ class RoverConfig(ModelConfig):
 
     """
 
+    model_type: Literal["binomial", "gaussian", "poisson"] | None = None
+    observation_column: str | None = None
+    weights_column: str | None = None
+    train_column: str | None = None
+    holdout_columns: set[str] | None = None
+    coef_bounds: dict[str, tuple[float, float]] | None = None
     cov_exploring: set[str]
```
blsmxiu47, line 78: `cov_exploring` seems not to be listed as explicitly required but has no default here. Is this fine? I guess attributes to Config classes are not the same as config settings, but just want to confirm

src/onemod/config/model_config/rover_config.py
```diff
@@ -49,14 +83,13 @@ class RoverConfig(ModelConfig):
...
-    #     """Make sure holdouts present."""
-    #     if self.holdout_columns is None:
-    #         raise ValueError("Holdout columns required for rover stage")
-    #     return self
+    _pipeline_config: Config = Config()
+    _required: set[str] = {
+        "model_type",
+        "observation_column",
+        "weights_column",
+        "holdout_column",
```
blsmxiu47, line 91: Is `holdout_column` distinct from `holdout_columns`? I guess I'm misunderstanding the relationship between `_required` items and the class attributes.

blsmxiu47: For context, trying to run `requires_data` tests on the cluster on this branch I get `AttributeError: Missing required config items: ['observation_column', 'weights_column', 'holdout_column']`, although I saw you have updated the test `DummyRoverStage` on this branch.

src/onemod/config/model_config/rover_config.py
kels271828: Oh, that's old! I'll remove it

src/onemod/config/model_config/rover_config.py
kels271828: So there are two kinds of required attributes: more general settings that can be inherited from the pipeline that have a `None` default and are included in `_required` (`add_pipeline_config()` will throw an error if not included in either stage or pipeline config), and stage-specific settings without a default (Pydantic will throw an error when initializing stage instance if not included)

We could always add the stage-specific attributes to `_required` to be less confusing, though the check in `add_pipeline_config()` would be unnecessary for them because Pydantic won't let them be unset. Open to other ways of handling these if we come up with other ideas!

src/onemod/config/model_config/rover_config.py
kels271828: That's a typo! I'm surprised the integration tests for model config passed...

kels271828: Thanks for catching this, my tests weren't written correctly and didn't catch the typo, but now should be fixed.

src/onemod/config/model_config/rover_config.py
blsmxiu47: Gotcha, I don't think it will be a problem so long as the docstrings + out-of-code documentation make sense, and I don't have better ideas at the moment

blsmxiu47:
If you don't mind checking the one commit I pushed to this branch, I still needed to add some dummy column names for the newly `_required` config fields to the instances of `RoverConfig` and `SpxmodConfig` in the test `DummyPipeline` setup in order to get all tests passing, and everything turned green so decided just to push.

At some point maybe we want to add tests with actual data (like testing failure when config columns are specified but they don't existing in the dataset), but since this would require reading of data objects it sounds like runtime validation which is a bigger task.

kels271828:
> If you don't mind checking the one commit I pushed to this branch, I still needed to add some dummy column names for the newly `_required` config fields to the instances of `RoverConfig` and `SpxmodConfig` in the test `DummyPipeline` setup in order to get all tests passing, and everything turned green so decided just to push.
> 
> At some point maybe we want to add tests with actual data (like testing failure when config columns are specified but they don't existing in the dataset), but since this would require reading of data objects it sounds like runtime validation which is a bigger task.

Looks good to me!

For data, would we use the input validation to check config columns?

blsmxiu47:
> > If you don't mind checking the one commit I pushed to this branch, I still needed to add some dummy column names for the newly `_required` config fields to the instances of `RoverConfig` and `SpxmodConfig` in the test `DummyPipeline` setup in order to get all tests passing, and everything turned green so decided just to push.
> > At some point maybe we want to add tests with actual data (like testing failure when config columns are specified but they don't existing in the dataset), but since this would require reading of data objects it sounds like runtime validation which is a bigger task.
> 
> Looks good to me!
> 
> For data, would we use the input validation to check config columns?

I think it could happen at the same time, but we probably want to refactor the validation logic. This would be a step of validation that happens regardless of whether the pipeline creator specifies `input_validation` or not. Don't know whether it should be optional.. (If not optional, that means there's always an extra read in of the data, and if using `pandas` that means reading in the entire dataset from file, which takes time. This is my main concern with runtime validation).

But anyway, if we were to add it it's probably 1) read in the data, 2) always check the Stage's `required` fields exist, plus any other validation we want to add, and then 3) while we've got the data if `input_validation` specs also happen to be specified, then perform validation for those as well.

blsmxiu47:
Also I would approve this but I think I forfeited my right to approve by making a commit on this PR so I think we also need @zhengp0 approval to merge sorry!

zhengp0 (approved):
Thanks very much @kels271828 and @blsmxiu47, this looks good to me!

PR #125, opened 2024-12-30
Replace set by UniqueList

blsmxiu47:
## Description

Addresses [MSCA-386](https://jira.ihme.washington.edu/browse/MSCA-368). Replaces instances of `set[str]` for class attributes and arguments with a new `UniqueList[str]` custom type hint. A `UniqueList` is ultimately a Python `List` (order matters and is preserved) but enforces items' uniqueness on initialization.

## Changes made

- Adds `UniqueList` custom type hint using `AfterValidator` for validation. By request, non-unique input is coerced rather than raising an error.
- Updates type hints (and default values) for class attributes and method arguments from `set[str]` to `UniqueList[str]`, where appropriate

## Additional notes

- non-jobmon tests all passing on cluster and local


zhengp0 (approved):
Looks great! Thanks very much @blsmxiu47 !

src/onemod/dtypes/unique_list.py
```diff
@@ -0,0 +1,25 @@
+from typing import Hashable, List, TypeVar
+
+from pydantic import AfterValidator
+from typing_extensions import Annotated
+
+T = TypeVar("T", bound=Hashable)
+
+
+def unique_list(items: List[T]) -> List[T]:
```
kels271828, line 9: Similar to `update_unique_list`, could `unique_list` be `return list(dict.from_keys(items))`?

kels271828:
omg looks like pydantic used to have a unique list (https://docs.pydantic.dev/latest/api/types/#pydantic.types.conlist) but got rid of it in this new version

kels271828:
Since `UniqueList` only behaves differently from `list` when creating class instances or calling a method with the `validate_call` decorator, maybe we don't need `UniqueList` type hints in methods that don't have the `validate_call` decorator. 

Also, I caught one instance where the return type was `UniqueList`, but there were actually duplicates (`Input.dependencies`). There may be other cases I didn't see.

src/onemod/stage/base.py
```diff
@@ -422,42 +422,42 @@ class ModelStage(Stage, ABC):
...
         return self._crossby
 
     @property
-    def subset_ids(self) -> set[int]:
+    def subset_ids(self) -> UniqueList[int | None]:
         if self.groupby is not None and not self._subset_ids:
             try:
                 subsets = self.dataif.load("subsets.csv", key="output")
-                self._subset_ids = set(subsets["subset_id"])
+                self._subset_ids = list(set(subsets["subset_id"]))
```
kels271828, line 440: Hadn't thought about it before, but maybe we want `subset_ids` and `param_ids` to be sorted?

src/onemod/stage/base.py
```diff
@@ -422,42 +422,42 @@ class ModelStage(Stage, ABC):
...
             except FileNotFoundError:
                 raise AttributeError(
                     f"{self.name} data subsets have not been created"
                 )
         return self._subset_ids
 
     @property
-    def param_ids(self) -> set[int]:
+    def param_ids(self) -> UniqueList[int | None]:
```
kels271828, line 448: Why the `None`?

tests/unit/io/test_input.py
```diff
@@ -223,13 +223,13 @@ def test_contains():
 @pytest.mark.unit
 def test_dependencies():
```
kels271828, line 224: We should add a test for duplicate dependencies.

kels271828:
Are there places where we've used `list(set(some_list))` where we should replace with `unique_list(some_list)` to preserve order?

blsmxiu47:
After implementing this my honest tl;dr thought is that it would be more explicit, less abstract, for us to pass around lists or sets, seeing as `UniqueList` is not really a type in the first place, and then validate the objects (e.g. for uniqueness) or sort them as/where it is needed. I think we were already doing this, at least in some places.

My perspective is that this pydantic abstraction does more harm than good in our situation, because as @kels271828 points out, there is no guarantee that anything will stay unique since these are all mutable objects and we are not overriding the `list` type. After the initial validation, it's just a `list`. As a contributor to the codebase I would rather accept the imperfections of Python's `list` (at least I know what behavior to expect) and explicitly check for and/or enforce uniqueness where it is required.

Not sure, is this perspective valid? There are surely ways to make better use of Pydantic validation that I'm not fully appreciating, but I'm not convinced we need more than basic Python types in this case.

zhengp0:
Indeed, it wasn't as clean as I expected. If we use `UniqueList` more than just `pydantic` validation, but also indicate some of the behaviors outside of `pydantic` it can be confusing and error prone. But I love the validation as part of the definition built into a class, this way we don't have to carry around extra validation function and remember to call it when we need it.

My preference is to create a `UniqueTuple` class (immutability make more sense) rather than just the type hint for `pydantic`. But I am also ok to move forward with just use `list` and validate when needed. And we can revisit this later. @blsmxiu47 @kels271828 let me know what you guys think!

kels271828:
I agree with @blsmxiu47 ! Maybe this `UniqueList` as implemented is not the way to go. I regret using sets in the first place, because it has lead us down this path. Either we create an actual type that behaves as an ordered set / unique list, or we just pass lists and check for uniqueness where necessary. The latter sounds cleaner!

kels271828:
@zhengp0 there are many cases where we need mutability. Maybe we have to think about each case where we were using a set, because a tuple is fine for some (e.g., groupby), but not for others.

zhengp0:
> @zhengp0 there are many cases where we need mutability. Maybe we have to think about each case where we were using a set, because a tuple is fine for some (e.g., groupby), but not for others.


Here is a very rough pass of the `UniqueTuple`, for the mutability part, we can use the `|` operand and return a new tuple.

```python
from typing import Self, Iterable, Generator, NoReturn, SupportsIndex, Any
from itertools import filterfalse, chain


def unique_everseen(iterable: Iterable, seen: tuple = ()) -> Generator:
    seen = set(seen)
    for item in filterfalse(seen.__contains__, iterable):
        seen.add(item)
        yield item

class UniqueTuple(tuple):

    def __new__(cls, iterable: Iterable = ()) -> Self:
        return super().__new__(cls, unique_everseen(iterable))

    def __or__(self, value: Iterable) -> Self:
        value = tuple(unique_everseen(value, seen=self))
        return super().__new__(type(self), chain.from_iterable((self, value)))
    
    def __add__(self, value: Any) -> NoReturn:
        raise NotImplementedError(
            f"unsupported operand type(s) for +: '{type(self).__name__}' and '{type(value).__name__}'"
        )

    def __mul__(self, value: SupportsIndex) -> NoReturn:
        raise NotImplementedError(
            f"unsupported operand type(s) for *: '{type(self).__name__}' and '{type(value).__name__}'"
        )

a = UniqueTuple([1, 1, 2, 3])
a | [1, 2, 4] # this will return a to be (1, 2, 3, 4)
```
Let me know what you guys think :)

kels271828:
This seems more complicated than necessary when all we needed were some checks for uniqueness in a few places

zhengp0:
Yea, have to give it a try :)  Then let's just pass in as list and check/convert uniqueness where needed.

blsmxiu47:
@zhengp0 Referencing the sub-classing of `tuple` - I was originally thinking the same thing, and embarked on a similar journey a week ago sub-classing `List`, so that we would be able to pass around something that is in fact a distinct type and behaves just like a regular `list` except every time you create a new one or try to mutate (e.g. calling `unique_list1.append(["a", "b", "c"])` or `unique_list1.extend([1, 2, 3])` or `unique_list1 + unique_list2` or `unique_list1 * 2`, etc etc), it would do a validation check to ensure you are not introducing duplicate values.

I learned a few things from this:
1) In order to override cleanly and completely, there a ton of methods to override and many edge cases to consider.
2) I felt the need to add dozens of unit tests to address the different situations where duplicates could be introduced, and even then don't think I managed to cover everything.
2) Dealing with `mypy` errors from of overriding `list` methods was going to be a huge pain.
4) At the end of the day it would be more to maintain, more for new contributors to try to understand, and probably would end up allowing for more room for human error than when just using lists + explicit validation because of the added complexity.

blsmxiu47:
> Yea, have to give it a try :) Then let's just pass in as list and check/convert uniqueness where needed.

Sounds like we are fine with this approach then. I think it will be best to keep simple until it becomes unwieldy and a need for more complexity becomes apparent.

Of the variables mentioned in the comment on [MSCA-356](https://jira.ihme.washington.edu/browse/MSCA-356), do we want to keep any as `Set` (as opposed to `List`), like `subset_ids` and/or `param_ids`? I wouldn't want to add in additional sorting operations where we don't need them.

kels271828:
> Of the variables mentioned in the comment on [MSCA-356](https://jira.ihme.washington.edu/browse/MSCA-356), do we want to keep any as `Set` (as opposed to `List`), like `subset_ids` and/or `param_ids`? I wouldn't want to add in additional sorting operations where we don't need them.

Both `subset_ids` and `param_ids` are created internally (with functions from `utils.subsets`) and never modified. I think I made them a set just to indicate that they were unique, but they are fine as a list or a tuple. I think based on how they are assigned, they will already be sorted if we never cast them as a set.

blsmxiu47:
Okay, so updated to more of a hybrid approach.

Replaced `some_func() -> UniqueList[...]` type hints with just `List[...]`, since there is no automatic validation on return type hints and this could be misleading. Instead manually include `unique_list(some_list)` wrapped around the list to be returned where we want to be sure we are returning a list without duplicates.

However left in `UniqueList[...]` type hints on `BaseModel` attributes and functions/methods with `@validate_call` decorator, since in these specific cases providing the type hint actually does mean that `unique_list()` gets applied on init.

Also added a few more tests related to this for peace of mind.

Please let me know if this approach seems reasonable, or if it's still too abstract.

PR #127, opened 2025-01-04
Merge Stage and ModelStage classes

kels271828:
WHAT
Merged `ModelStage` class into `Stage` class.

WHY
Difference between stages, especially `ModelStage` name, were a bit confusing/misleading. Now all stages can use `groupby`/`crossby` (but don't have to; previously `ModelStage` had to use one or the other), and all stages must explicitly define behavior for all `fit`/`predict` methods (previously `Stage` classes only had to implement `run`). Note: It's not assumed that a particular stage does fitting or predicting (e.g., preprocessing, plotting, cleanup, etc.), it's that we can call `fit` and `predict` on the entire pipeline, and we need to define how each stage will behave when these methods are called.

HOW
- Moved all `ModelStage` functions into `Stage` class, added `has_submodels` property to easily check if stage can run in parallel, removed `ModelStage` class
- Updated local and Jobmon backends
- Updated subset and parameter set creation in `pipeline.build()`
- Updated relevant tests

### DIFF
--- a/src/onemod/backend/local_backend.py
+++ b/src/onemod/backend/local_backend.py
@@ -82,13 +82,13 @@ def _evaluate_stage(
         if stage.has_submodels:
             eval_submodel = False
             if subset_id is not None:
-                if stage.groupby is None:
+                if len(stage.groupby) == 0:
                     raise ValueError(
                         f"Stage '{stage.name}' does not use groupby attribute"
                     )
                 eval_submodel = True
             if param_id is not None:
-                if stage.crossby is None:
+                if len(stage.crossby) == 0:
                     raise ValueError(
                         f"Stage '{stage.name}' does not use crossby attribute"
                     )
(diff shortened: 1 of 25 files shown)


src/onemod/backend/local_backend.py
```diff
@@ -62,24 +62,35 @@ def _evaluate_stage(
...
-            stage.collect()
-        else:
-            if subset_id is None and param_id is None:
+    if method == "collect":
+        stage.collect()
+    else:
+        stage_method = stage.__getattribute__(method)
+        if stage.has_submodels:
+            subset_passed = stage.groupby is not None and subset_id is not None
+            param_passed = stage.crossby is not None and param_id is not None
```
zhengp0, line 84: Here I think maybe we should error out if `stage.groupby is None and subset_id is not None` same with the `crossby`

kels271828: Good idea

blsmxiu47:
There was one e2e test that I needed to update locally on this branch, not counting the newer `backend` tests. Just with how the `DummyPreprocessingStage` logs are checked, but it seems to be fixed by https://github.com/ihmeuw-msca/OneMod/pull/128. Going to approve these so that we're not blocking you and will be more scrutinous (scrutinitious?) on the request to merge `tests/jobmon` into `release/1.0`

blsmxiu47 (approved):

kels271828:
> There was one e2e test that I needed to update locally on this branch, not counting the newer `backend` tests. Just with how the `DummyPreprocessingStage` logs are checked, but it seems to be fixed by #128. Going to approve these so that we're not blocking you and will be more scrutinous (scrutinitious?) on the request to merge `tests/jobmon` into `release/1.0`

Sounds good!

PR #128, opened 2025-01-05
Update crossby attribute

kels271828:
WHAT
Update `crossby` attribute to work the same way as `groupby`.

WHY
- Declaring parameters to parallelize over via the `crossby` attribute (e.g., `crossby=["param"]`) is more explicit than just having `param` as a list of parameters in the stage config.
- Users won't have to create their own custom stage config with `_crossable_params`.

HOW
- Added `stage.submodel_ids` generator to fix a mypy error
- Removed `StageConfig._crossable_params` private attribute
- Changed `groupby`, `crossby`, `subset_ids`, and `param_ids` to tuples, changed default from None to an empty tuple
- Added `field_validator` for `Pipeline.groupby`, `Stage.groupby`, and `Stage.crossby` to enforce unique items
- Added `crossby` field to `Stage` and removed private attribute `_crossby`
- Added `model_post_init` methods to `RoverStage` and `SPxModStage` to require `groupby` attribute but prevent `crossby`
- Updated `subsets` and `parameters` utility modules for tuples
- Updated relevant tests

QUESTIONS
- I updated `groupby`, `crossby`, `subset_ids`, and `param_ids` to be tuples instead of sets. May have some merge conflicts when merge with the `UniqueList` PR
- Relevant to both `groupby` and `crossby`: Do we want a way to allow users to prevent the use of either attribute?
  - If pipeline has `groupby`, it will get passed to all stages. But we could have a stage (e.g., preprocessing) that we don't want to have `groupby`. Right now, the thing to do would be to NOT have `groupby` in pipeline, and then any column that we want all parallel stages to have would have to be included in each of the parallel stage's definitions. What if someone could say `groupby=False` instead? Note: Unless we do something like this, it is always possible for a stage to have subsets, so all functions (except `collect`) will need to have `subset_id` as an argument.
  - Since I removed `_crossable_params`, a user can parallelize over any parameter in a stage's config. Do we want to restrict which parameters can be parallelized over, or restrict a stage from using `crossby` altogether? For `RoverStage` and `SPxModStage`, I added a check in `model_post_init` that does not allow `crossby`, but there could be a better way. Unless a stage does something like this, it will always be possible for a stage to have parameter sets, so all functions (except `collect`) will need to have `param_id` as an argument.

### DIFF
--- a/src/onemod/stage/base.py
+++ b/src/onemod/stage/base.py
@@ -7,10 +7,16 @@ from abc import ABC, abstractmethod
 from functools import cached_property
 from inspect import getfile
 from pathlib import Path
-from typing import Any, Literal
+from typing import Any, Generator, Literal
 
 from pandas import DataFrame
-from pydantic import BaseModel, ConfigDict, Field, validate_call
+from pydantic import (
+    BaseModel,
+    ConfigDict,
+    Field,
+    field_validator,
+    validate_call,
+)
 
 import onemod.stage as onemod_stages
 from onemod.config import StageConfig
@@ -46,10 +52,12 @@ class Stage(BaseModel, ABC):
         Stage name.
     config : StageConfig
         Stage configuration.
-    groupby : set of str or None, optional
-        Column names used to create data subsets. Default is None.
-    crossby : set of str or None, optional
-        Parameter names used to create parameter sets. Default is None.
+    groupby : tuple of str, optional
+        Column names used to create data subsets. Default is an empty
+        tuple.
+    crossby : tuple of str, optional
+        Parameter names used to create parameter sets. Default is an
+        empty tuple.
     input_validation : dict, optional
         Optional specification of input data validation.
     output_validation : dict, optional
@@ -90,8 +98,8 @@ class Stage(BaseModel, ABC):
 
     name: str
     config: StageConfig
-    groupby: set[str] | None = None
-    crossby: set[str] | None = None
+    groupby: tuple[str, ...] = tuple()
+    crossby: tuple[str, ...] = tuple()
     input_validation: dict[str, Data] = Field(default_factory=dict)
     output_validation: dict[str, Data] = Field(default_factory=dict)
     _dataif: DataInterface | None = None
@@ -102,8 +110,16 @@ class Stage(BaseModel, ABC):
     _output: set[str] = set()
     _skip: set[str] = set()
     _collect_after: set[str] = set()
-    _subset_ids: set[int] = set()
-    _param_ids: set[int] = set()
+    _subset_ids: tuple[int, ...] = tuple()
+    _param_ids: tuple[int, ...] = tuple()
+
+    @field_validator("groupby", "crossby", model="after")
+    @classmethod
+    def unique_tuple(cls, items: tuple[str, ...]) -> tuple[str, ...]:
+        """Make sure groupby and crossby have unique values."""
+        if len(items) > 0:
+            return tuple(dict.fromkeys(items))
+        return items
 
     @property
     def dataif(self) -> DataInterface:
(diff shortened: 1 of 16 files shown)


src/onemod/stage/base.py
```diff
@@ -215,85 +230,104 @@ def dependencies(self) -> set[str]:
...
-
     @property
     def has_submodels(self) -> bool:
-        return self.groupby is not None or self.crossby is not None
+        return len(self.groupby) > 0 or len(self.crossby) > 0
 
     @property
-    def subset_ids(self) -> set[int]:
-        if self.groupby is not None and not self._subset_ids:
+    def submodel_ids(self) -> Generator:
```
blsmxiu47, line 238: Interesting. So will this be the standard way to access `subset_ids` and `param_ids` together, i.e. you need to write a loop or comprehension or something to consume the `Generator` right?

I see it's only used in one place at the moment but I guess I would wonder if we'd ever want a way to return a static object that contains all the submodel ids? Possibly not. It's just that the `subset_ids` and `param_ids` properties are static tuples. Not sure if that discrepancy matters though.

src/onemod/stage/base.py
```diff
@@ -215,85 +230,104 @@ def dependencies(self) -> set[str]:
...
-            columns=list(self.groupby),
-            id_subsets=id_subsets,
-            return_type="pandas_dataframe",
+            key=data_key, columns=list(self.groupby), id_subsets=id_subsets
         )
 
         subsets_df = create_subsets(self.groupby, df)
-        self._subset_ids = set(subsets_df["subset_id"].to_list())
-
+        self._subset_ids = tuple(subsets_df["subset_id"])
```
blsmxiu47, line 291: I do like this better thanks

src/onemod/utils/parameters.py
```diff
@@ -8,25 +8,29 @@
...
-        param_name: param_values
-        for param_name in config.crossable_params
-        if isinstance(param_values := config[param_name], set)
-    }
-    if not param_dict:
-        return None
+    param_dict = {}
+    for param_name in crossby:
+        param_values = config[param_name]
+        if isinstance(param_values, (list, set, tuple)):
```
blsmxiu47, line 16: Just to confirm, is the `(list, set, tuple)` because we are fine with any of them going forward, or because we haven't decided a single standard yet? If the latter, let's make sure to update this.

src/onemod/utils/subsets.py
```diff
@@ -3,16 +3,17 @@
...
 
 
-def create_subsets(groupby: set[str], data: pd.DataFrame) -> pd.DataFrame:
+def create_subsets(
+    groupby: tuple[str, ...], data: pd.DataFrame
+) -> pd.DataFrame:
     """Create subsets from groupby."""
-    sorted_groupby = sorted(groupby)
-    groups = data.groupby(sorted_groupby)
+    groups = data.groupby(list(groupby))
```
blsmxiu47, line 10: Think we briefly talked about this before but I guess this is maybe a downside(?) of using `tuple`, i.e. we need to convert to `list` anyway for some fields. I think I still prefer tuple whenever we want the thing to be immutable for non-zero amount of its life though

tests/e2e/test_e2e_dummy_pipeline_sequential.py
```diff
@@ -54,12 +54,7 @@ def test_dummy_pipeline(small_input_data, test_base_dir, method):
     expected_args = get_expected_args()
 
     for stage in stages:
-        if stage.name == "preprocessing":
-            if method in ["run", "fit"]:
-                assert stage.get_log() == [f"run: name={stage.name}"]
-            else:
-                assert stage.get_log() == []
-        elif stage.name in expected_args:
+        if stage.name in expected_args:
```
blsmxiu47, line 57: This fixes an issue in https://github.com/ihmeuw-msca/OneMod/pull/127

blsmxiu47 (approved):

src/onemod/stage/base.py
kels271828: I think this change was actually because of a mypy error (I can't remember the error exactly, but there are probably other ways to fix it than this solution). Can definitely replace with something static as well. I think we might need to rethink this when I change subset_ids to be more interpretable anyway.

src/onemod/utils/parameters.py
kels271828: I think we can replace it with something like `Sequence`. The type will depend on what's in the config class, so it could be any list, set, tuple, etc.

src/onemod/utils/subsets.py
kels271828: One alternative is to make `groupby`/`crossby` fields frozen lists. I think I started going that route, but one problem was that the stage `groupby` gets modified by the pipeline `groupby` in `build()`. We could get rid of `pipeline.groupby` altogether. Downside would be that if user wants all stages to groupby `sex_id`, they have to add it to every stage. Upside is that it might be a little more transparent when reading the stage definition, so I'm not opposed to this.

PR #129, opened 2025-01-24
Interpretable subsets/paramsets

kels271828:
**Description**
For stages that use the `groupby`/`crossby` features, we previously used integer IDs `subset_id`/`param_id` to handle submodel data subsets/parameter sets. However, integer IDs were inconvenient because we frequently had to look up the ID corresponding to different demographic groups. We changed the integer IDs to dictionaries called `subset`/`paramset` because dictionaries allow us to immediately see which demographic group or parameter values a submodel corresponds to.

**Unchanged behavior**
- Pipeline methods `run`/`fit`/`predict` always run all submodels
- Options to run all submodels:
  - Command line:
    - All pipeline stages: `onemod config=/path/to/config.json method=run`
    - Subset of pipeline stages: `onemod config=/path/to/config.json method=run stages=[first_stage, second_stage, ...]`
    - Individual stage: `onemod config=/path/to/config.json method=run stages=stage`
  - Script/notebook:
    - All pipeline stages: `pipeline.run()`
    - Subset of pipeline stages: `pipeline.run(method=run, stages=[first_stage, second_stage, ...])`

**Previous behavior**
- Stage attributes `subset_ids`/`param_ids` contain lists of integers
- Options to run a single submodel:
  - Command line: `onemod config=/path/to/config.json method=run stages=stage subset_id=0 param_id=0`
  - Script/notebook: `stage.run(subset_id=0, param_id=0)`

**New behavior**
- Stage attributes `subsets`/`paramsets` stage attributes contain data frames
- Options to run a single submodel:
  - Command line: `onemod config=/path/to/config.json method=run stages=stage subsets="{'sex_id': 1, 'location_id': 1}" paramsets="{'param1': 1}"`
  - Script/notebook: `stage.run(subset={"sex_id": 1, "location_id": 1}, paramset={"param": 1})`
- (New!) Options to run subset of submodels (e.g., single sex, all locations, subset of param values):
  - Command line: `onemod config=/path/to/config.json method=run stages=stage subsets="{'sex_id': 1}" paramsets="{'param': [1, 2]}"`
  - Script/notebook: `stage.run(subsets={sex_id: 1}, paramsets={"param": [1, 2]})`
- Options to run all submodels:
  - Command line: `onemod config=/path/to/config.json method=run stages=stage`
  - Script/notebook: `stage.run()` 
  
**Other changes**
- Removed `pipeline.groupby` attribute. Stages that don't use submodels (e.g., preprocessing) no longer have to worry about handling unused `groupby` passed from `pipeline` (this wasn't an issue before we merged `Stage` and `ModelStage`). When defining stages that do use submodels, you must include all columns in their own `groupby`, which is more explicit anyway. Added `cov_groupby` to `RoverConfig` to indicate how to group covariates (previously used `pipeline.groupby`).
- Removed `pipeline.id_subsets` attribute. If users want to run entire pipeline on subset of data, they just need to filter their data outside of the pipeline. `id_subsets` could only include columns from `pipeline.groupby` anyway, which limited how useful the feature was. Allowing columns outside of `pipeline.groupby` was overly complicated when filtering data subsets, so decided it wasn't worth the fuss.
- Removed `utils.subsets` and `utils.parameters` modules. The change from integer IDs to dictionaries simplified a lot of data subset/parameter set interactions, so these modules were no longer needed.
- Updated `RoverStage`, `SpxmodStage`, and relevant tests for `subsets`/`paramsets` changes
- Updates to `fsutils`: change `id_subsets: dict[str, list[Any]]` to `subsets: dict[str, Any | list[Any]]` in `data_loader`, update `path_manager` to load files from paths not associated with a key, change `id_subsets` to `subsets` in `interface`
- Added optional `collect` arg to stage methods to control whether or not submodel results are collected when you aren't running all of the submodels (default is `False`)
- User now has to implement `_run`/`_fit`/`_predict` methods for to evaluate a single submodel for custom stages (`run`/`fit`/`predict` now passes to backend)
- Added optional `methods_args: dict` arg to allow args in addition to `subset/paramset` in custom stage `_run`/`_fit`/`_predict` (specifically thinking for subnational case where we might want to pass path to different train/test data)
- Added checks in `backend.utils`
- Pulled in changes from UniqueList PR


kels271828:
@saalUW This API change will require updates to the codem pipelines. Let me know if you have any comments/suggestions!

src/onemod/backend/local_backend.py
```diff
@@ -1,105 +1,92 @@
...
+        If None, evaluate all pipeline stages. Default is None.
+    subsets : dict, optional
+        Submodel data subsets to evaluate if `model` is a `Stage`
+        instance. If None, evaluate all data subsets. Default is None.
+    paramsets : dict, optional
+        Submodel parameter sets to evaluate if `model` is a `Stage`
+        instance. If None, evaluate all parameter sets. Default is None.
+    collect : bool, optional
+        Collect submodel results if `model` is a `Stage` instance and
+        `subsets` and `params` are not both None. Default is False.
```
blsmxiu47, line 42:
```suggestion
        `subsets` and `paramsets` are not both None. Default is False.
```

src/onemod/pipeline.py
```diff
@@ -283,48 +283,25 @@ def save_validation_report(
...
-            Path to resources file or dictionary of compute resources.
-            Required if `backend` is 'jobmon'.
-
-        """
+        """Evaluate pipeline method."""
         if backend == "jobmon":
             from onemod.backend.jobmon_backend import evaluate
         else:
-            from onemod.backend.local_backend import evaluate
+            from onemod.backend.local_backend import evaluate  # type: ignore
```
blsmxiu47, line 297: Just curious do you remember what the issue was which required the `# type: ignore`?

src/onemod/stage/base.py
```diff
@@ -520,27 +492,214 @@ def get_field(
...
+        paramsets : dict, optional
+            Submodel parameter sets to evaluate. If None, evaluate all
+            data subsets. Default is None.
+        collect : bool, optional
+            Collect submodel results if `subsets` and `paramsets` are
+            not both None. Default is False. If `subsets` and
+            `paramsets` are not both None, this parameter is ignored and
+            submodel results are always collected.
+
+        Other Parameters
```
blsmxiu47, line 522: Mere thoughts here - Referring to the `Other Parameters`, this is probably a necessary side effect of passing the `**kwargs` through multiple levels. Of course definitely prefer to have this info in the docstring than not have it, it's just looking at this docstring (and the others like it) in isolation took me a second to process for some reason (maybe because there are `**kwargs` that might be required arguments based on `backend` value, if I understand correctly). If users of the package don't interact with this directly it's totally fine. And even if they did, I don't necessarily have a better solution; as long as we have clear documentation for the pipeline/stage-building aspect of OneMod, all good.

And thank you for including detailed docstrings. Makes it way easier to understand as well as identify potential areas of confusion.

src/onemod/backend/local_backend.py
kels271828: Thanks!

src/onemod/pipeline.py
kels271828: It was a mypy error due to the two backend `evaluate` functions having different args (the jobmon backend has `cluster`, `resources`, `and python`):

```
src/onemod/pipeline.py: note: In member "evaluate" of class "Pipeline":
src/onemod/pipeline.py:297: error: Incompatible import of "evaluate" (imported name has type "Callable[[Pipeline | Stage, Literal['run', 'fit', 'predict', 'collect'], dict[str, Any | dict[str, Any]] | None, list[str] | None, dict[str, Any | list[Any]] | None, dict[str, Any | list[Any]] | None, bool], None]", local name has type "Callable[[Pipeline | Stage, Literal['run', 'fit', 'predict', 'collect'], str, dict[str, Any] | Path | str, Path | str | None, dict[str, Any | dict[str, Any]] | None, list[str] | None, dict[str, Any | list[Any]] | None, dict[str, Any | list[Any]] | None, bool], None]")  [assignment]
Found 1 error in 1 file (checked 1 source file)
```

src/onemod/stage/base.py
kels271828: Yeah, I wasn't sure which things to add in Parameters vs. Other Parameters. I put most things in `kwargs` to make things shorter, but maybe it's better to have everything there (it will be more explicit and allows type checking by pydantic).

blsmxiu47:
Thanks for the responses. Once again I can't seem to approve the PR but Approved from my perspective~

src/onemod/stage/base.py
kels271828: I updated methods by adding all arguments, denoting whether they were pipeline/stage/jobmon-specific parameters in docstrings where relevant. Hopefully this will make it clearer!

src/onemod/pipeline.py
kels271828: Ended up renaming the backend `evaluate` functions to be backend-specific, so no longer needed the ignore comment

PR #130, opened 2025-02-05
Developer Guide docs updates for release/1.0

blsmxiu47:
### Description

Updating/Adding "Developer Guide" docs prior to `release/1.0`.

### Changes made

* Update and clean up developer guide, especially testing tips
* rm outdated section of repo readme (the replacement pipeline usage guide will go into `docs/` anyway)
* Added a "OneMod Support" page. Not sure what other channels of communication we might want to list, but we can add to this as appropriate.
* Minor custom css adjustments

### Additional notes

* RE: custom css styles - Having issue locally with furo theme, so still need to test how it looks with furo. Not the most urgent thing but probably will update styles one more time before release


kels271828 (approved):

blsmxiu47:
> Looks good! I didn't know about the `requires_jobmon` marker (maybe you just created it). The tests I have right now run on a dummy cluster so should be able to run with GitHub actions, though the jobmon e2e tests take forever on my laptop for some reason.

I created it a while back but it probably hasn't been used. If we can run jobmon tests with GitHub Actions that's great(!) and I can amend the wording of the docs.

PR #131, opened 2025-02-07
Added tests for local and Jobmon backends

kels271828:
There are lots of files change in this PR because I had chained feature branches off of this branch (won't do that again). The files to look at for this PR are:
- backend/jobmon_backend.py
    - Incorporate new functions from backend/utils.py
    - Re-order functions
    - Pass all `op_args`, `task_args`, and `node_args` to `create_task()`/`create_tasks()` instead of putting them directly into the command template with an f-string (in particular, jobmon had trouble if stage kwargs included a dictionary, but this change also made some tests easier to write)
- backend/local_backend.py
    - Incorporate new functions from backend/utils.py
- (new) backend/utils.py
    - Move some checks and utility functions from pipeline/stage/backends
- pipeline.py
    - Moved some checks in `evaluate()` into backend/utils.py
- stage/base.py
    - Moved some checks in `evaluate()` into backend/utils.py
- tests/conftest.py
    - Added simple/parallel pipelines and resources directory as fixtures
- (new) tests/helpers/orchestration_helpers.py
    - Created two pipelines to test pipeline orchestration
    - `setup_simple_pipeline()` returns a pipeline without submodels, can also be use to test `Stage.skip` and method-specific dependencies (once implemented)
    - `setup_parallel_pipeline()` returns a pipeline with submodels, can also be used to test `Stage.collect_after`
    - `assert_{simple/parallel}_logs()` and `assert_{simple/parallel}_output()` functions check stage logs/outputs
- (new) tests/e2e/test_e2e_jobmon_backend.py
- (new) tests/e2e/test_e2e_local_backend.py
- (new) tests/integration/test_integration_jobmon_backend.py
- (new) tests/unit/backend/test_backend_utils.py
- (new) tests/unit/backend/test_jobmon_backend.py

Note: The Jobmon e2e tests take FOREVER to run (not sure why).


tests/unit/utils/test_subsets.py
```diff
@@ -1,74 +1,76 @@
-import pandas as pd
-import pytest
+# FIXME: Removed subsets module from utils, move to stage integration tests
```
blsmxiu47, line 1: If we cannot easily move this now, maybe make a ticket?

blsmxiu47 (approved):
Thank you Kelsey! Incredible amount of value added here.

Currently waiting for the jobmon e2e tests to run, but assuming the tests all pass I have just a couple of items:

1. Commented in a couple places but I think if we can make Jira tickets in lieu of `TODO` or `FIXME` comments from `release/1.0` forward this will make task management easier going forward while keeping the code clean. Personally I will treat the ticket queue as the SSOT for OneMod tasks/priority, whereas in my experience often TODOs in the codebase are liable to remain unaddressed for a long time.
2. Jobmon tests - You mentioned that we could include jobmon tests in CI. I am not 100% sure where the line is (i.e. which tests to include if not all of them.. I would be concerned about adding tests that take a very long time to run), but perhaps we remove the `requires_jobmon` and `requires_data` markers from being filtered from the CI workflow and instead add a marker that is like `"exclude_from_ci"` or similar. This doesn't have to be done now, can make a ticket for it. Plus related but different task - some tests I added a while back with `requires_data` could easily be using a tiny mock dataset small enough to include with version control, or even construct in-line with `pandas`, and could also be included in the CI tests. This can all be post-release ofc.
3. The time required for e2e testing - fine for now, but I want to look into this further when I'm back in office, if you wouldn't mind making a ticket. For me it seems to be taking a very very long time, as you mentioned, to the point that I know in the future I may end up doing less TDD (test-driven development) on the jobmon backend because I dread having to wait for the tests to run.

Thank you again! 🎉

src/onemod/backend/jobmon_backend.py
```diff
@@ -25,280 +34,642 @@
         {cluster_name}:
             ...
 
-See Jobmon documentation for additional optional resources and default
-values.
+See Jobmon documentation for additional resources and default values.
 
 """
+# TODO: Optional stage-specific Python environments
```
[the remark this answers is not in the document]
kels271828: Agree! We might need to make a ticket to go through the code and find all of the todo comments.

tests/unit/utils/test_subsets.py
kels271828: I can make a ticket now, but might be able to get to it after finishing documentation this week.

PR #132, opened 2025-02-14
User Guide

kels271828:
**Doc changes:**
Adding documentation to the User Guide
* user_guide/index.rst
* user_guide/core_concepts.rst
* user_guide/advanced_usage.rst

**Code changes:**
As I was writing documentation for creating custom stages, I realized we needed to update how we specified stage input/output (e.g., for future method-related dependencies). Since this would be an API change, I wanted to do it before the release (otherwise I wouldn't have done it in this docs branch). This lead to some other changes/simplifications. The docstrings need cleaned up, but that will be a separate docs PR.
* dtypes/data.py
  * Replaced `DataIOHandler` with `dataio_dict` and `configio_dict` in `validate_metadata` method
  * Replaced `DataIOHandler` with `DataLoader` in `validate_data` method
  * Added `methods` attribute to `Data` class for future method-related dependencies
  * Added `model_post_init` method to check that `path` attribute's suffix matches `format` attribute
* fsutils/interface.py
  * Added default for `key` to `load` and `dump` methods
* fsutils/path_manager.py
  * Added default for `key` in `get_full_path` method
* io/base.py
  * Removed option for `items` to contain paths, now all `Input` / `Output` items must be instances of `Data` (this simplifies a lot of methods)
  * `Input` class:
    * Change `required` and `optional` attributes from `list[str]` to `dict[str, Data]`. This makes input specification clearer and allows the class to include additional validation requirements (e.g., `"data": {"format": "parquet", **options}}` rather than "data.parquet"), and it also makes `_expected_names` and `expected_types`  private attributes unnecessary
    * Add `collector` arg to `check_missing` and `check_exist` methods
    * Added `path_to_data` method
  * `Output` class:
    * Added `directory` attribute to `Output` class
    * Added `model_post_init` method to add `stage` and `path` attributes to items
    * Added `check_exists` method
* pipeline.py
  * Reordered methods
  * Removed `model_post_init` method (directory now created in `build` method)
  * Add pipeline `config` to stage `config` attribute, initialize stage `dataif` attribute, and initialize stage `output` attribute in `add_stage` method (this allows stage output items to have full paths rather than just name.extension)
  * Moved some code from pipeline's `build` method to a new stage `build` method
* stage/base.py
  * Added `__init__` method with optional `config_path`, `input`, and `module` args (including `config_path` means stage's `dataif` can be created before adding stage to pipeline, the `input` and `module` args simplify the `from_json` method)
* model_stages (kreg_stage.py, rover_stage.py, spxmod_stage.py)
  * Updated format of `_required_input`, `_optional_input`, and `_output_items` private attributes (**Question**: I chose dict`[str, dict]`, but we could do `dict[str, Data]` instead. Opinions?)
  * Added `set_module`, `set_input`, and `set_output` methods rather than awkwardly creating them within the property methods
  * Simplify `set_dataif` method now that stage output items must be `Data` objects with full paths
  * Simplify `from_json` method now that we have more args in new `__init__` method
  * Add `build` method
  * Add validation methods to `evaluate` method (will need to work on `validate_output`; needs to be method-specific)
  * Added call to `input.check_missing` in `validate_build` method
  * Added call to `input.check_exists` in `validate_run` and `validate_output` methods
  * Update `dataif` when defining dependencies in `__call__` method, remove call to `input.check_missing` (this now happens in `build`)
* validation/error_handling.py
  * Add option for `stage` arg to be `None` (for cases where stage input comes from a path rather than an upstream stage)
* tests
  * Update tests for changes

### DIFF
--- a/docs/user_guide/advanced_usage.rst
+++ b/docs/user_guide/advanced_usage.rst
@@ -1,48 +1,35 @@
-.. _advanced_usage:
-
 Advanced Usage
 ==============
 
+Data
+----
+The :py:class:`~onemod.dtypes.Data` class contains information about input /
+output items such as which stage and method they correspond to, the path where
+the item will be saved, and other metadata (e.g., format, shape, columns, etc.)
+used for validation.
+
 Input / Output
 --------------
-The :py:class:`~onemod.io.base.Input` and
-:py:class:`~onemod.io.base.Output` classes are dictionary-like objects
-that help organize stage dependencies and paths to stage input/output
-items. Each stage has an :py:attr:`~onemod.stage.base.Stage.input` and
-an :py:attr:`~onemod.stage.base.Stage.output` attribute. Input items,
-either paths or upstream stage output, are added to the stage's `input`
-attribute when defining pipeline dataflow using the stage's
-:py:meth:`~onemod.stage.base.Stage.__call__` method. The `Input` object
-checks that all required items have been defined and that all items have
-the correct type. Output items, instances of
-:py:class:`~onemod.dtypes.data.Data`, are added to the stage's `output`
-attribute automatically. These objects contain information about the
-output item such as which stage created the item, the path where the
-item will be saved, and other metadata (e.g., shape, columns, etc.) used
-for validation. The stage's `output` attribute is returned when invoking
-the stage's :py:meth:`~onemod.stage.base.Stage.__call__` method when
-defining pipeline dataflow:
-
-.. code-block:: python
-
-   preprocessing_output = preprocessing_stage(raw_data="/path/to/raw_data.parquet")
-   modeling_output = modeling_stage(observations=preprocessing_output["modeling_data"])
-   plotting_output = plotting_stage(
-      observations=preprocessing_output["plotting_data"],
-      predictions=modeling_output["predictions"],
-   )
+The :py:class:`~onemod.io.base.Input` and :py:class:`~onemod.io.base.Output`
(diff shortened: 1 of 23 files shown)


zhengp0 (approved):
Thanks very much @kels271828! I just took a brief look and everything looks great! One minor comment is about the `Data` class. I understand it is designed to be a descriptor of the data. But I think it might be good to also include the name of the file. Currently we need to work with a dictionary like `dict[str, Data]` in the stages to complete the information, but if user provide the `path` in `Data` we actually have the file name information from the path. I think this confused me a bit on what to expect from `Data` and if we work with `dict[str, Data]` what is the `str` stands for. Sorry for the long message! We can address this in the future PR :)

kels271828:
> Thanks very much @kels271828! I just took a brief look and everything looks great! One minor comment is about the `Data` class. I understand it is designed to be a descriptor of the data. But I think it might be good to also include the name of the file. Currently we need to work with a dictionary like `dict[str, Data]` in the stages to complete the information, but if user provide the `path` in `Data` we actually have the file name information from the path. I think this confused me a bit on what to expect from `Data` and if we work with `dict[str, Data]` what is the `str` stands for. Sorry for the long message! We can address this in the future PR :)

I don't think the user ever interacts with the `Data` objects, but we definitely need more documentation about what the `str` stands for (it corresponds to the keys to `Stage.__call__()` and `Stage.dataif`, plus the file name in `Stage.output`). We can discuss this in a meeting to figure out what's the most intuitive.

PR #133, opened 2025-02-14
API Reference Documentation

kels271828:
Getting the basic API documentation organized and written. So far I just have `main` (command line entrypoint), `Pipeline`, and `Stage` published, and they all could use some more details and examples.

Let me know if any other modules should be documented before the release, otherwise we will continue to add documentation later.


zhengp0 (approved):
Thanks @kels271828! These look good!

PR #138, opened 2025-02-25
Bugfix: dependencies, stage init, and model stages

kels271828:
Changes
- added deptry to check dependencies
- removed examples, model stages, and model configs (we will probably add these back in the future)
- pass stage fields as kwargs in stage.__init__ (this allows custom stages to have field default values)


kels271828:
Technically this should be a "hotfix" not a "bugfix"

blsmxiu47 (approved):
Tested that this builds locally for me. Approving so that we can test deploying this sooner rather than later

PR #141, opened 2025-02-28
Hotfix: Input class serializer

kels271828:
Changes to `Input` class:
- Model serializers now only returns a dictionary, instead of a dictionary or None
- Added `exclude_none=True` arg so that empty `Data` fields don't get serialized

Changes to `Stage` class:
- Added `OSError` to the catch statement in `Stage.set_module()`(was previously only catching `TypeError`, but I managed to get an `OSError` while I was debugging the `Input` issue

Note: If we want to go the other route of more control over the pipeline serialization, we can cancel this PR and move forward with Sameer's branch!

### DIFF
--- a/src/onemod/io/base.py
+++ b/src/onemod/io/base.py
@@ -38,7 +38,9 @@ class IO(BaseModel, ABC):
         # Simplify output to config files
         input_dict: dict[str, dict[str, Any]] = {}
         for item_name, item_value in self.items.items():
-            input_dict[item_name] = item_value.model_dump(serialize_as_any=True)
+            input_dict[item_name] = item_value.model_dump(
+                exclude_none=True, serialize_as_any=True
+            )
         return input_dict
 
     def get(self, item_name: str, default: Any = None) -> Any:
--- a/src/onemod/io/base.py
+++ b/src/onemod/io/base.py
@@ -34,16 +34,12 @@ class IO(BaseModel, ABC):
     items: dict[str, Data] = {}
 
     @model_serializer
-    def serialize_io(self) -> dict[str, dict[str, Any]] | None:
+    def serialize_io(self) -> dict[str, dict[str, Any]]:
         # Simplify output to config files
-        if self.items:
-            input_dict: dict[str, dict[str, Any]] = {}
-            for item_name, item_value in self.items.items():
-                input_dict[item_name] = item_value.model_dump(
-                    serialize_as_any=True
-                )
-            return input_dict
-        return None
+        input_dict: dict[str, dict[str, Any]] = {}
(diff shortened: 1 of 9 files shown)


src/onemod/io/base.py
```diff
@@ -34,16 +34,14 @@ class IO(BaseModel, ABC):
     items: dict[str, Data] = {}
 
     @model_serializer
-    def serialize_io(self) -> dict[str, dict[str, Any]] | None:
+    def serialize_io(self) -> dict[str, dict[str, Any]]:
```
blsmxiu47, line 37: So this will remove `None` items from the serialized output, correct? I feel like at some point we decided to keep them to differentiate between the value being `None` and the key-value pair itself missing (which might indicate a bug or something). Has this perspective changed? If so, I think there are a couple of tests we need to update.

Failing test assertion diff from `tests/unit/io/test_input.py::test_serialize`:
```
Full diff:
    {
        'covariates': {
  -         'columns': None,
            'format': 'csv',
  -         'methods': None,
            'path': PosixPath('/path/to/selected_covs.csv'),
  -         'shape': None,
            'stage': 'first_stage',
        },
        'data': {
  -         'columns': None,
            'format': 'parquet',
  -         'methods': None,
            'path': PosixPath('/path/to/predictions.parquet'),
  -         'shape': None,
  -         'stage': None,
        },
        'priors': {
  -         'columns': None,
            'format': 'pkl',
  -         'methods': None,
            'path': PosixPath('/path/to/model.pkl'),
  -         'shape': None,
            'stage': 'second_stage',
        },
    }
```

blsmxiu47 (changes requested):
Couple of unit tests fail for me with this update. So we either need to find a slightly different strategy or else update the test cases themselves if the proposed change is desirable.

I agree that how we handle serialization probably does deserve a deeper look eventually, but sounds like something we could think more about and integrate into the regular release cycle. If this fix works then I think we go through with this for the hotfix and @saalUW if you want to build upon your branch toward a longer term solution (as your priorities allow, etc.) we can plan on adding that to the 1.1 cycle!

src/onemod/io/base.py
kels271828: An alternative to the fix I have now, which wouldn't affect the current serialization, would be to update `Stage.set_input()` to ignore any input values other than `Input` or dict (i.e., the 'null' value that's getting saved). I could make this change and we could handle serialization later when we can discuss how we really want things to behave.

kels271828: Yes, I can definitely do this. I completely forgot to run tests when I created the PR, so thanks for catching the failing tests!

PR #143, opened 2025-03-12
Patch: Require jobmon_installer_ihme>=10.9.0

blsmxiu47:
## Description

Require `jobmon_installer_ihme>=10.9.0` for our `jobmon` dependency to be extra safe in prep for the upcoming Slurm 24.11 upgrade: https://slurm.schedmd.com/release_notes.html

## Notes

OneMod `pytest` tests pass with `10.9.0` and there should not be any changes needed to OneMod API itself, but this makes it clear that if running into any Jobmon backend orchestration issues, the first step is to ensure the jobmon version is up-to-date. Again, just an abundance of caution.


kels271828 (approved):

kels271828:
Kelly pointed out that onemod 1.0.3 already exists on PyPI, how did that happen when there isn't a 1.0.3 tag on GitHub?

saalUW (approved):

PR #147, opened 2025-03-17
Feature/jobmon workflow passthrough

lzoeckle:
Updating Jobmon backend to accept a new workflow argument and other arguments related to that new functionality.

New arguments include:

1. workflow: allows the backend to run on an existing Jobmon Tool/Workflow, rather than creating a new Tool/Workflow. Additionally, if passed, do not run the Workflow as part of execution; just add the tasks.
2. external_upstream_tasks: a list of Jobmon tasks to be added as upstreams to every task in the OneMod Pipeline/Stage(s).
3. task_and_template_prefix: a string to append to the front of every task and task template added in the OneMod Pipeline/Stage(s). This allows multiple OneMod Pipelines to exist in the same Workflow without name collisions.
4. max_attempts: an integer, this just surfaces the Jobmon max_attempts arg (which was previously hard-coded).


blsmxiu47:
Thanks @lzoeckle!

Not sure if there's enough context in the existing OneMod user docs to add example(s) of using this new feature. For instance maybe in the "Advanced Concepts" section we add something that gives an example of using an to existing Jobmon wf. But I realize we don't really have user documentation for the jobmon backend in the first place to build on so not sure I can ask you to do that 😅. If we want to merge in this functionality asap maybe we need to address the user documentation separately.. Or else we could add a some simple example(s) of how to run with jobmon, existing workflow and not, into this feature PR.

And I suppose, while I wait for tests to run..., this is the first case for OneMod where we have a direct feature branch into `main`, so I imagine this will be `v1.1.0` - We'll need to update `CHANGELOG.md`, update the version strings in `pyproject.toml` and `meta.toml`, and then the existing `release/1.1` branch will become `release/1.2`. @kels271828 or @zhengp0 does this sound right? I am operating under the assumption that we want to include this feature in `main` sooner rather than later. Also let me know if I'm forgetting anything.

src/onemod/backend/jobmon_backend.py
```diff
@@ -434,11 +463,22 @@ def get_stage_tasks(
         **kwargs,
```
kcercy, line 463: Does `task_and_template_prefix` get passed through to get_task_template with kwargs?

src/onemod/backend/jobmon_backend.py
```diff
@@ -381,6 +398,9 @@ def get_stage_tasks(
     paramsets: dict[str, Any | list[Any]] | None = None,
     collect: bool | None = None,
     upstream_tasks: list[Task] | None = None,
+    external_upstream_tasks: list[Task] | None = None,
+    task_and_template_prefix: str | None = None,
```
kcercy, line 402: All of the get_stage_tasks callsites looks like they have kwargs except for the recursive call for collect in this function. That's why none of these arguments need to be explicitly passed through? They'll be kwargs to the top-level `evaluate_with_jobmon` and get used as needed from there?

kels271828: Some context: `kwargs` is meant to be any stage `run`/`fit`/`predict` method keyword arguments other than `subsets` and/or `paramsets`. The stage `collect` method currently can't have any keyword arguments, which is why we don't pass them when we create the `collect` task. However, regarding the new arguments:
- Technically we don't need to pass `external_upstream_tasks` to the `collect` task because the `collect` task's upstream tasks will already have have them in their upstream tasks. However, if you wanted to be consistent you could pass them.
- We do need to pass `task_and_template_prefix` to the `collect` task

kels271828:
I think instead of modifying `evaluate_with_jobmon`, we could get the functionality you want with something like this: `workflow.add_tasks(get_tasks(pipeline, method, workflow.tool, ...))`. The main reason I suggest this is that the name `evaluate_with_jobmon` indicates that you are _evaluating_ a pipeline, so it's a little confusing to call the function but not evaluate anything. Same with how `pipeline.evaluate(workflow=workflow, ...)` is called in the new e2e test (at first glance, it looks like you are evaluating the pipeline, not just adding the pipeline's tasks to the workflow).

kels271828:
I think it could be helpful to make the new args `external_upstream_tasks`, `task_and_template_prefix`, and `max_attempts` explicit and not included in `**kwargs` for some functions but not others. The original intention was that `**kwargs` corresponded to keyword arguments for the stage `run`/`fit`/`predict` other than `subsets` and/or `paramsets`.

kels271828:
This might be overkill, but we don't technically need to add the `external_upstream_tasks` to every task: 
- In `get_pipeline_tasks()`, we could add them to the `task_dict` (i.e., `task_dict['external'] = external_upstream_tasks`). Then in `get_upstream_tasks()`, if `upstream_tasks` was empty and `external in task_dict`, we could return `task_dict['external']`.
- In `get_stage_tasks()` (i.e., if you were adding a single stage to your workflow rather than the whole pipeline), you could just pass `upstream_tasks=external_upstream_tasks` rather than having two separate arguments.

src/onemod/backend/jobmon_backend.py
lzoeckle: Yes!

lzoeckle:
> I think it could be helpful to make the new args `external_upstream_tasks`, `task_and_template_prefix`, and `max_attempts` explicit and not included in `**kwargs` for some functions but not others. The original intention was that `**kwargs` corresponded to keyword arguments for the stage `run`/`fit`/`predict` other than `subsets` and/or `paramsets`.

Hmmm is that not what's happening now? They're kwargs right now for anywhere they're not used explicitly

lzoeckle:
Ok @kels271828 I think I addressed all your comments! Let me know if you see anything else

src/onemod/backend/jobmon_backend.py
```diff
@@ -257,17 +360,28 @@ def get_tasks(
...
+            tool,
+            resources,
+            python,
+            stages,
+            external_upstream_tasks,
+            task_and_template_prefix,
+            max_attempts,
+            **kwargs,
         )
     return get_stage_tasks(
```
kels271828, line 374: I think we also need to pass `upstream_tasks=external_upstream_tasks` to `get_stage_tasks()`

src/onemod/pipeline.py
```diff
@@ -260,6 +263,28 @@ def save_validation_report(
         report_path = validation_dir / "validation_report.json"
         serialize(collector.errors, report_path)  # type: ignore[arg-type]
 
+    def add_tasks_to_workflow(
```
kels271828, line 266: Other people can chime in, but since we tried to keep the orchestration separate from the pipeline/stage classes, I'm wondering if this function belongs in the jobmon backend rather than the pipeline class.

src/onemod/__init__.py
```diff
@@ -1,9 +1,11 @@
+from onemod.backend.jobmon_backend import add_tasks_to_workflow
```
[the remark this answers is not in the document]
zhengp0: Yes, might be good to explicitly import from `onemod.backend.jobmon_backend`

zhengp0:
Thanks @lzoeckle for the PR! Here are some general comments and ideas :)

**1. This might need work on our end, but it might be beneficial to separate out workflow creation and running.**
This can leave further manipulations of the Jobmon workflow (if needed) open as well. For example we can have
```python
def create_workflow(
    model: Pipeline | Stage,
    method: Literal["run", "fit", "predict", "collect"],
    cluster: str,
    resources: Path | str | dict[str, Any],
    python: Path | str | None = None,
    stages: list[str] | None = None,
    subsets: dict[str, Any | list[Any]] | None = None,
    paramsets: dict[str, Any | list[Any]] | None = None,
    collect: bool | None = None,
    task_and_template_prefix: str | None = None,
    max_attempts: int = 1,
    **kwargs,
) -> Workflow: ...

def run_workflow(workflow: Workflow) -> None: ...
```
@kels271828 @blsmxiu47 Let me know what you guys think!

**2. Extract the post manipulation of the workflow as functions outside of the workflow creation**
To be honest I don't know if this is possible, but it will be nice if the post process of the workflow doesn't have to worry about the implementation details of the workflow creation itself. For example
```python
def prepend_external_tasks(workflow: Workflow, external_tasks: list[Task]) -> Workflow: ...
```
This will rely on if we can parse out which tasks in the workflow are the root of the tree (no dependency) and add `external_tasks` to their upstream. But I don't know if this is feasible.
@lzoeckle @blsmxiu47 @kcercy let me know what you guys think!

kels271828:
> **2. Extract the post manipulation of the workflow as functions outside of the workflow creation** To be honest I don't know if this is possible, but it will be nice if the post process of the workflow doesn't have to worry about the implementation details of the workflow creation itself. For example
> 
> ```python
> def prepend_external_tasks(workflow: Workflow, external_tasks: list[Task]) -> Workflow: ...
> ```
> 
> This will rely on if we can parse out which tasks in the workflow are the root of the tree (no dependency) and add `external_tasks` to their upstream. But I don't know if this is feasible. @lzoeckle @blsmxiu47 @kcercy let me know what you guys think!

I think they need the onemod tasks to belong to the external workflow, which is different from just adding the external tasks as dependencies in a separate onemod workflow

kels271828:
> **1. This might need work on our end, but it might be beneficial to separate out workflow creation and running.** This can leave further manipulations of the Jobmon workflow (if needed) open as well. For example we can have
> 
> ```python
> def create_workflow(
>     model: Pipeline | Stage,
>     method: Literal["run", "fit", "predict", "collect"],
>     cluster: str,
>     resources: Path | str | dict[str, Any],
>     python: Path | str | None = None,
>     stages: list[str] | None = None,
>     subsets: dict[str, Any | list[Any]] | None = None,
>     paramsets: dict[str, Any | list[Any]] | None = None,
>     collect: bool | None = None,
>     task_and_template_prefix: str | None = None,
>     max_attempts: int = 1,
>     **kwargs,
> ) -> Workflow: ...
> 
> def run_workflow(workflow: Workflow) -> None: ...
> ```

This wouldn't be too hard to do, would we add it to this PR or v1.2?

zhengp0:
> This wouldn't be too hard to do, would we add it to this PR or v1.2?

Great! Thanks @kels271828! I am thinking maybe we can add to this PR, but want to make sure this is helpful for @lzoeckle and @kcercy

zhengp0:
> I think they need the onemod tasks to belong to the external workflow, which is different from just adding the external tasks as dependencies in a separate onemod workflow

I see, that makes sense, then the current `add_tasks_to_workflow` makes more sense.

lzoeckle:
> > This wouldn't be too hard to do, would we add it to this PR or v1.2?
> 
> Great! Thanks @kels271828! I am thinking maybe we can add to this PR, but want to make sure this is helpful for @lzoeckle and @kcercy

I don't think we'll use this/need this, but it should be an easy change! Happy to make it as part of this PR.

lzoeckle:
> > This wouldn't be too hard to do, would we add it to this PR or v1.2?
> 
> Great! Thanks @kels271828! I am thinking maybe we can add to this PR, but want to make sure this is helpful for @lzoeckle and @kcercy

Actually, I may have misunderstood this: would you like these to be methods on Pipeline? Still easy, just making sure I understand!

kels271828:
> > > This wouldn't be too hard to do, would we add it to this PR or v1.2?
> > 
> > 
> > Great! Thanks @kels271828! I am thinking maybe we can add to this PR, but want to make sure this is helpful for @lzoeckle and @kcercy
> 
> Actually, I may have misunderstood this: would you like these to be methods on Pipeline? Still easy, just making sure I understand!

I think we want to keep any jobmon-related code in the jobmon backend, so none of these methods, including `add_tasks_to_workflow` should be in `Pipeline`. 

My thought is something like this:

`
def evaluate_with_jobmon(
    model: Pipeline | Stage,
    method: Literal["run", "fit", "predict", "collect"],
    cluster: str,
    resources: Path | str | dict[str, Any],
    python: Path | str | None = None,
    stages: list[str] | None = None,
    subsets: dict[str, Any | list[Any]] | None = None,
    paramsets: dict[str, Any | list[Any]] | None = None,
    collect: bool | None = None,
    prefix: str | None = None,
    max_attempts: int = 1,
    **kwargs,
) -> None:
    resources_dict = get_resources(resources)
    workflow = create_workflow(model, method, cluster, resources_dict, prefix)
    add_tasks_to_workflow(
        model,
        workflow,
        method,
        resources_dict,
        python,
        stages,
        subsets,
        paramsets,
        collect,
        prefix,
        max_attempts,
        **kwargs,
    )
    run_workflow(workflow)


def create_workflow(
    model: Pipeline | Stage,
    method: Literal["run", "fit", "predict", "collect"],
    cluster: str,
    resources: dict[str, Any],
    prefix: str | None = None,
) -> Workflow:
    tool = get_tool(model, method, cluster, resources, prefix)
    workflow = tool.create_workflow(name=get_name(model, method, prefix))
    return workflow


def get_tool(
    model: Pipeline | Stage,
    method: str,
    cluster: str,
    resources: dict[str, Any],
    prefix: str | None = None,
) -> Tool:
    tool = Tool(name=get_name(model, method, prefix))
    tool.set_default_cluster_name(cluster)
    tool.set_default_compute_resources_from_dict(
        cluster, resources["tool_resources"][cluster]
    )
    return tool


def get_name(
    model: Pipeline | Stage, method: str, prefix: str | None = None
) -> str:
    name = f"{model.name}_{method}"
    if prefix is None:
        return name
    return f"{prefix}_{name}"


def run_workflow(workflow: Workflow) -> None:
    workflow.bind()
    print(f"Starting workflow {workflow.workflow_id}")
    status = workflow.run()
    if status != "D":
        raise ValueError(f"Workflow {workflow.workflow_id} failed")
    else:
        print(f"Workflow {workflow.workflow_id} finished")
`

zhengp0:
Thanks @kels271828! `create_workflow` and `run_workflow` look good to me. @lzoeckle I think once the import of `jobmon_backend` is removed from `onemod.__init__` we are good to go with this PR :)

kels271828 (approved):
Thanks @lzoeckle ! It looks good to merge.

PR #148, opened 2025-03-19
Hotfix: Issues with custom pipelines

kels271828:
WHAT
Fixed issues related to custom `Pipeline` and `Config` classes when reloading from JSON files.

WHY
* Custom pipeline objects were being reloaded as the base `Pipeline` class; would be missing any custom fields or methods and config would not be validated
* Custom pipeline config objects were being reloaded as base `Config` class in the stage config; pipeline config fields would not be validated
  * Noticed by Peng when running a stage on Jobmon. In particular, a pipeline config item was not converted from str to Path when loaded from JSON file. 

HOW
* `pipeline.py`:
  * Added `type` and `module` attributes to `Pipeline`
  * Added `__init__()` function to `Pipeline`
* `main.py`:
  * Generalized `_get_stage()` and `_get_custom_stage()` methods to work with stages and pipelines; renamed to `_get_class()` and `_get_custom_class()`
  * Updated `load_pipeline` to use `_get_class()` function rather than returning a base `Pipeline` object
* `config.py`: Infer type of `_pipeline_config` rather than using a base `Config` object
* `advanced_usage.rst`: Updated documentation related to custom `Config` and `StageConfig` classes

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 # OneMod
 
 **OneMod** is an orchestration package that allows users to build pipelines of
-various models created by [IHME Math Sciences](https://github.com/ihmeuw-msca).
+statistical models created by [IHME Math Sciences](https://github.com/ihmeuw-msca).
 Core features of **OneMod** include an intuitive syntax for defining the
 dataflow between pipeline stages, the ability to easily parallelize over
 different data subsets and/or parameter sets, and options for data validation.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -10,7 +10,7 @@ Welcome to OneMod!
    developer_guide/index
 
 **OneMod** is an orchestration package that allows users to build pipelines of
-various models created by `IHME Math Sciences <https://github.com/ihmeuw-msca>`_.
+statistical models created by `IHME Math Sciences <https://github.com/ihmeuw-msca>`_.
 Core features of **OneMod** include an intuitive syntax for defining the
 dataflow between pipeline stages, the ability to easily parallelize over
 different data subsets and/or parameter sets, and options for data validation.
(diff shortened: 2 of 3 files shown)


kels271828:
I wasn't sure where this fix would go (i.e., 1.0.4 vs. 1.1.1), so I haven't updated the version/changelog/etc.

src/onemod/main.py
```diff
@@ -74,34 +79,46 @@ def _get_stage(config: Path | str, stage_name: str) -> Stage:
...
+def _get_custom_class(
+    class_type: str, module: str
+) -> type[Pipeline] | type[Stage]:
+    """Get custom pipeline or stage class from file.
 
     Parameters
     ----------
-    stage_type : str
-        Name of custom stage class.
+    class_type : str
```
blsmxiu47, line 113: Might just be tripping myself up over nomenclature. I know `model_type` (the result of `config_dict["type"]`) above is passed for this param but the wording of `class_type` might be confusing me. Is this param just the class name?

kels271828: Yeah, would it be better to change to `class_name`?

zhengp0: `class_name` makes sense to me :)

src/onemod/pipeline.py
```diff
@@ -40,8 +41,42 @@ class Pipeline(BaseModel):
...
+        **kwargs,
+    ) -> None:
+        """Create pipeline instance."""
+        super().__init__(**kwargs)
+        self.set_module(module)
+        if stages is not None:
+            self.add_stages(stages)
+
+    @computed_property
+    def type(self) -> str:
```
zhengp0, line 60: Should we change `type` to `class_name` as well?

kels271828: I prefer keeping `type` here

blsmxiu47 (approved):

PR #150, opened 2025-04-04
Fix how `Config` and `StageConfig` handle None values

kels271828:
`Config` and `StageConfig` no longer ignore None-valued items. If a `StageConfig` item has the same name as a pipeline `Config` item, the `get` and `__getitem__` methods will always return the stage's item even if it is None (updated behavior). In this case, the only way to get the pipeline's item is to use the `get_from_pipeline` method.

See MSCA-440 for more background.


zhengp0 (approved):
Thanks very much Kelsey for quickly making the PR!! This looks good to me

PR #156, opened 2025-04-14
Splitting task and template prefixes

lzoeckle:
Splits task_and_template_prefix into task_prefix and template_prefix, so we can use the same template across all models in a given WF


tests/e2e/test_e2e_jobmon_backend.py
```diff
@@ -159,3 +188,32 @@ def test_parallel_pipeline_add_tasks_to_workflow(parallel_pipeline):
     )
     workflow.bind()
     workflow.run()
+
+
+@pytest.mark.skip(reason="Bug in Jobmon dummy cluster/null.q execution.")
```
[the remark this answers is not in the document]
blsmxiu47: Pulled down these changes and ran this test on its own (dummy cluster, null.q) and it passed 🤔. Definitely make a ticket if you can with environment and we'll take a deeper look. @kels271828 would be curious if you have time to try running this test locally whether you also get an error.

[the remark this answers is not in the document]
lzoeckle: Oh interesting! What python version?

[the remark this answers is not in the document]
blsmxiu47: 3.10.16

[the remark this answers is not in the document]
blsmxiu47: I'll test for 3.10~3.12 (what we claim to support)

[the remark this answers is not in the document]
lzoeckle: Rebuilding the env did it! I must have gotten in a messed up state somehow, I'll remove this skip

src/onemod/backend/jobmon_backend.py
```diff
@@ -765,17 +804,21 @@ def get_task_template(
...
+    # OneMod models in a single workflow; since each model will have
+    # a separate config, this allows a user to keep the same task
+    # template name across all models in a given workflow.
     task_template = tool.get_task_template(
         template_name=template_name,
         command_template=get_command_template(method, submodel_args, **kwargs),
         op_args=["entrypoint"],
-        task_args=["config", "method", "stages"] + list(kwargs.keys()),
-        node_args=submodel_args,
+        task_args=["method", "stages"] + list(kwargs.keys()),
```
kels271828, line 820: Could this potentially get messed up if someone passed optional kwargs sometimes but not other times? I don't anticipate this would ever happen for some onemod_cod runs and not others (to my knowledge no pipelines currently have additional kwargs to run/fit/predict/collect), so would probably never come up.

kels271828 (approved):

src/onemod/backend/jobmon_backend.py
lzoeckle: Yes, I was a little worried about this! I'm not sure there's a good way to solve this, other than to require that if you're passing different kwargs per model, you need to pass a template_prefix as well as a task_prefix?

kcercy: I might not be following the problem! kwargs were task_args previously, that hasn't changed in the scope of this PR, right? Is the question if they should switch over to node_args along with config? That if someone were to pass kwargs that determined stage parallelization, this would trigger a jobmon error about duplicate tasks?

lzoeckle: I think it's more that in the previous version, if you were adding multiple models to a single WF they necessarily used multiple templates. Now, they could use the same template, but in theory they could have different kwargs! In which case the template definition would be different (the contents of kwargs would be different), despite having the same name. FWIW I think the same problem exists with submodel_args as well, the definition of the template could change despite having the same name.

kcercy: Got it! So one jobmon workflow might have tasks across multiple verisons (definitions) of the same task template. My instinct is to leave it as is, and assume if the user does that it's itentional? It looks like jobmon can handle it, I just built a fake workflow with two defitions and it validated okay. Agree it's not likely to come up with the known use cases.

kcercy (approved):

src/onemod/backend/jobmon_backend.py
lzoeckle: Doesn't the second one get overwritten by the first? I just put up a kind of gnarly check, but would love to not have to use it!

src/onemod/backend/jobmon_backend.py
```diff
@@ -765,17 +804,40 @@ def get_task_template(
...
     """
     template_name = (
-        f"{task_and_template_prefix}_{stage_name}_{method}"
-        if task_and_template_prefix
+        f"{template_prefix}_{stage_name}_{method}"
+        if template_prefix
         else f"{stage_name}_{method}"
     )
 
+    command_template = get_command_template(method, submodel_args, **kwargs)
```
lzoeckle, line 812: Added this fairly gnarly check for existing task templates.

@blsmxiu47 do you know how jobmon handles two templates with identical names but different command templates/task args/node args?

lzoeckle: Nevermind, removing this! @kcercy 's check looks good to me, and it makes it seem like the current functionality is safe/works. No need for this check.

PR #157, opened 2025-05-05
Adds distinct upstreams for different task/template prefixes, and adds task attributes

lzoeckle:
The new functionality that allowed multiple models in a single workflow also introduced a bug, where all subsequent models would depend on the previous model's corresponding upstream stages to complete before launching their individual downstream stages.

As an example, with two models, A and B, model B's fit_location_run tasks would depend on both model A _and_ B's fit_national_collect task, instead of just model B's.

This PR updates the upstream logic to only look for upstreams that share task and template prefixes, if those prefixes exist.

### DIFF
--- a/src/onemod/backend/jobmon_backend.py
+++ b/src/onemod/backend/jobmon_backend.py
@@ -75,7 +75,7 @@ def evaluate_with_jobmon(
     paramsets: dict[str, Any | list[Any]] | None = None,
     collect: bool | None = None,
     task_prefix: str | None = None,
-    task_attributes: dict[str, Any] | None = None,
+    task_attributes: dict[str, Any] = dict(),
     template_prefix: str | None = None,
     max_attempts: int = 1,
     **kwargs,
@@ -123,10 +123,10 @@ def evaluate_with_jobmon(
     task_prefix : str, optional
         Optional prefix to append to task names. Default is None, no
         prefix.
-    task_attributes : dict, optional
-        Optional dictionary containing task attribute names and values.
-        Note that the task attributes will be shared across all tasks in
-        any given Workflow. Default is None, no task attributes.
+    task_attributes : dict
+        Dictionary containing task attribute names and values. Note that
+        the task attributes will be shared across all tasks in any given
+        Workflow. Default is an empty dict, no task attributes.
     template_prefix : str, optional
         Optional prefix to append to task template name. Default is None,
         no prefix.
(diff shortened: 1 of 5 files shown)


lzoeckle:
Closing this for now, I think I may want to also add task attributes here

lzoeckle:
Re-opening, now with task attributes included (and a higher version bump to account for the new arg)

src/onemod/backend/jobmon_backend.py
```diff
@@ -557,15 +590,31 @@ def get_upstream_tasks(
...
 
         upstream_stage = stage_dict[upstream_name]
+
+        # Filter possible upstreams using task and template prefixes
+        possible_upstream_tasks = task_dict[upstream_name]
+        if task_prefix:
+            possible_upstream_tasks = [
+                possible_upstream_task
+                for possible_upstream_task in possible_upstream_tasks
+                if task_prefix in possible_upstream_task.name
```
kcercy, line 600: Rather than checking `in`, could we tighten an use `.startswith()`? Same thought on checking the task template name.

src/onemod/backend/jobmon_backend.py
```diff
@@ -557,15 +590,31 @@ def get_upstream_tasks(
...
+            ]
+        if template_prefix:
+            possible_upstream_tasks = [
+                possible_upstream_task
+                for possible_upstream_task in possible_upstream_tasks
+                if template_prefix
+                in possible_upstream_task.node.task_template_version.task_template.template_name
```
kcercy, line 607: This is a mouthful! Whew

src/onemod/backend/jobmon_backend.py
```diff
@@ -557,15 +569,31 @@ def get_upstream_tasks(
...
+            ]
         if method not in upstream_stage.skip:
             if (
                 upstream_stage.has_submodels
                 and method in upstream_stage.collect_after
             ):
                 # only include task corresponding to 'collect' method
-                upstream_tasks.append(task_dict[upstream_name][-1])
+                upstream_tasks.append(possible_upstream_tasks[-1])
```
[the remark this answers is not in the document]
kcercy: We could do another list comprehension on `possible_upstream_tasks` to parse out the task(s) that end with "_collect"? Agree on currently functional though! May not be worth it.

kcercy (approved):

src/onemod/backend/jobmon_backend.py
lzoeckle: Nothing private, but yeah, it's...fairly hideous 😅

saalUW (approved):

PR #159, opened 2025-05-05
Remove MSCA File System Dependency

blsmxiu47:
## Description

Removed the dev requirement for setting a `TEST_ASSETS_DIR` (which pointed to a directory containing static test assets.. not robust), given that we are not currently making reasonable use of it. Would have been relevant if we were integration testing with very large datasets or something that needed to live in the file system, but we ended up not doing any of that so far.

## Changes made

* tests: replace reference to TEST_ASSETS_DIR env var for small test dataset with a dataset generated as part of pytest fixture. Should also allow for what used to be "requires_data"-marked tests to be run as a part of CI.
  * Leaving the pytest marker for now in case we do later have a need to exclude something that requires data from CI, but with this PR it currently applies to no tests. Can remove it entirely if requested.
* Moved `get_expected_args()` helper to a more logical home.
* deps: `python-dotenv` no longer needed as dev dependency.
* local env: Similarly, remove `.env.example` as it is no longer needed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,1 +1,0 @@
-TEST_ASSETS_DIR=
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,6 @@ dev = [
     "pre-commit>=4.0.1,<5.0.0",
     "pytest>=7.0.0",
     "pytest-cov>=5.0.0",
-    "python-dotenv",
     "ruff>=0.7.0,<1.0.0",
     "types-PyYAML",
 ]
@@ -57,7 +56,6 @@ jobmon = [
 test = [
     "pytest>=7.0.0",
     "pytest-cov>=5.0.0",
-    "python-dotenv",
 ]
 
 [project.urls]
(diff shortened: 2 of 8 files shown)


zhengp0 (approved):
Thanks @blsmxiu47!

blsmxiu47:
> Looks good to me, but isn't the current release 1.2, not 1.1? Apologies if that's an issue I created with my last PR!

Yes so this release cycle will actually be at least 1.3+ 😅 as a consequence of us not being focused on OneMod package development recently.

Originally I didn't want to rename the release branch directly since it can make the git history harder to track, though that might be the simpler approach. In general I would prefer to create a new branch (e.g. `release/1.2`, `release/1.3`) each time `main` gets a new minor version outside of the regular release cycle, but with focus elsewhere lately have fallen behind on that.

Unless there are other proposals/preferences for release branch management for OneMod, can PR to push all of `release/1.1` into a new `release/1.3` immediately after this.

PR #162, opened 2025-10-29
Pass columns and subset args through to read_parquet to optimize memory usage

kcercy:
**What:**  `columns` and `subsets` arguments to `fsutils.data_loader.DataLoader.load` are passed through as `options` to `load_eager` when `path` is a parquet file to optimize memory usage.

**Why:** I'm working through high memory usage in an application of this library that loads a dataset with many columns, one for each out-of-sample model prediction. To lower memory usage in steps that only care about in-sample, I'd like to skip reading the OOS columns. This can be done in the current API with the columns arg! But I wasn't fully convinced that this was passed all the way through to read_parquet, and instead what might be happening is the whole dataset is read into memory and columns are subset afterwards. 

The new test I wrote fails on main, so I've convinced myself that there's optimization to be had! What I'm not confident of is if this is the best way to go about it. This method is used pretty much everywhere so I'm wary of downstream impacts. Thanks for the review!

- [to do] run OneMod CoD off of this branch

### DIFF
--- a/src/onemod/fsutils/data_loader.py
+++ b/src/onemod/fsutils/data_loader.py
@@ -30,37 +30,17 @@ class DataLoader:
             raise ValueError(f"Unsupported data format for '{path.suffix}'")
 
         if return_type == "pandas_dataframe":
-            if path.suffix == ".parquet":
-                if columns:
-                    options["columns"] = columns
-                if subset and "filters" not in options:
-                    options["filters"] = [
-                        (
-                            col,
-                            "in",
-                            values if isinstance(values, list) else [values],
-                        )
-                        for col, values in subset.items()
-                    ]
-
-            pandas_df = self.io_dict[path.suffix].load_eager(path, **options)
+            pandas_df = self.io_dict[path.suffix].load_eager(
+                path, columns=columns, subset=subset, **options
+            )
             assert isinstance(pandas_df, pd.DataFrame), (
                 "Expected a pandas DataFrame"
             )
 
+            # Enforce `column` order if requested and reset index in case of `subset`.
             if columns:
                 pandas_df = pandas_df[columns]
-
-            if subset:
-                for col, values in subset.items():
-                    pandas_df = pandas_df[
-                        pandas_df[col].isin(
-                            values if isinstance(values, list) else [values]
-                        )
-                    ]
-                    pandas_df.reset_index(drop=True, inplace=True)
-
-            return pandas_df
+            return pandas_df.reset_index(drop=True)
         elif return_type in ["polars_dataframe", "polars_lazyframe"]:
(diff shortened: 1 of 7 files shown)


kels271828 (approved):

zhengp0 (approved):

src/onemod/fsutils/data_loader.py
```diff
@@ -30,24 +30,17 @@ def load(
...
-                pandas_df, pd.DataFrame
-            ), "Expected a pandas DataFrame"
+            pandas_df = self.io_dict[path.suffix].load_eager(
+                path, columns=columns, subset=subset, **options
+            )
+            assert isinstance(pandas_df, pd.DataFrame), (
+                "Expected a pandas DataFrame"
+            )
 
+            # Enforce `column` order if requested and reset index in case of `subset`.
```
kcercy: Keeping this to retain backwards compatibility, `columns` in read_parquet and `use_cols` in read_csv won't change the column order, only subset down to those columns.

src/onemod/fsutils/io.py
```diff
@@ -150,11 +200,35 @@ def _dump_pandas_impl(self, obj: pd.DataFrame, fpath: Path, **options):
...
-        return pl.read_parquet(fpath, **options)
+    def _load_eager_polars_impl(
+        self,
+        fpath: Path,
+        columns: list[str] | None = None,
+        subset: Mapping[str, Any | list[Any]] | None = None,
+        **options,
+    ) -> pl.DataFrame:
+        return _subset_dataframe(
+            pl.read_parquet(fpath, columns=columns, **options), subset=subset
```
kcercy: No polars option to filter rows on read, subsetting after the fact! Better practice seems to be scan_parquet -> select which is covered in the lazy implementation!

src/onemod/fsutils/data_loader.py
```diff
@@ -30,24 +30,17 @@ def load(
...
             if columns:
                 pandas_df = pandas_df[columns]
```
zhengp0: If we deal with the columns parameters in the detailed `_impl` function, I think we can skip the `columns` logic here.

src/onemod/fsutils/io.py
```diff
@@ -150,11 +200,35 @@ def _dump_pandas_impl(self, obj: pd.DataFrame, fpath: Path, **options):
...
+            [
+                (col, "in" if isinstance(values, list) else "=", values)
+                for col, values in subset.items()
+            ]
+            if subset
+            else None
+        )
+        return pd.read_parquet(
+            fpath, columns=columns, filters=filters, **options
+        )
```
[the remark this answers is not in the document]
kels271828: If someone happened to pass `filters` in `**options`, it looks like the `filters` created here might get ignored?

